PR #1, opened 2026-03-28
feat(ui): redesign layout and theme to M3-inspired design

chrisgen19:
## Summary
- Redesign entire UI to match a Material 3-inspired design system with crimson red (#c3172f) color palette
- Replace complex shadcn SidebarProvider with a lightweight custom sidebar (~130 lines vs ~720 lines)
- Add floating white content area with `rounded-l-[2rem]` on gray background
- Redesign auth pages, tables, forms, and dialogs with pill buttons, large rounded cards, taller inputs, and hover-reveal actions
- Full dark mode support across all new components

## Test plan
- [ ] Visit `/login` and `/register` — verify centered card with crimson theme, rounded inputs, pill buttons
- [ ] Visit `/dashboard` — verify sidebar with pill nav items, floating white content area
- [ ] Visit `/dashboard/users` — verify M3 table with search bar, avatar initials, hover-reveal actions
- [ ] Visit `/dashboard/users/new` and edit pages — verify M3 form styling
- [ ] Visit `/dashboard/departments` — verify M3 table, create/edit/delete dialogs
- [ ] Visit `/dashboard/profile` — verify M3 form styling
- [ ] Test mobile: sidebar hidden, hamburger menu triggers sheet sidebar
- [ ] Toggle dark mode — verify all components render correctly


chrisgen19:
Review findings:

- [P2] Restore non-hover access to staff row actions — `app/(dashboard)/dashboard/users/_components/user-list-client.tsx:178`
  On touch-only devices and for keyboard users, these row actions never become visible because the container stays at `opacity-0` unless the row is hovered. Since the redesign also removed the always-visible overflow menu and the row itself is not clickable, admins can't reliably view, edit, or activate/deactivate a user from the staff directory without a mouse.

- [P2] Restore non-hover access to department actions — `app/(dashboard)/dashboard/departments/_components/department-table.tsx:91`
  On touch devices and in keyboard-only navigation, the edit/delete buttons remain fully transparent because they are only revealed by `group-hover`. That makes the redesigned table's only department-management controls effectively inaccessible anywhere a hover state isn't available.

- [P3] Keep the department editor modal keyboard-dismissible — `app/(dashboard)/dashboard/departments/_components/department-form.tsx:78`
  For keyboard users opening this dialog, replacing the previous `Dialog` with a plain fixed `<div>` removes the built-in focus trap and `Esc` handling. Focus can now move behind the overlay, and there is no keyboard-only way to dismiss the editor, which is a regression in the admin CRUD flow.

- [P3] Align this staff column label with the data rendered — `app/(dashboard)/dashboard/users/_components/user-list-client.tsx:124`
  This header says `Role / Dept`, but the cells in this column render `user.position` and `user.department?.name`. Whenever someone's job title differs from their permission role, the table now presents inaccurate information to admins.

chrisgen19:
## Fixes Applied

All 4 review issues have been addressed in commit a73d464:

### [P2] Restore non-hover access to staff row actions
Row actions now use `sm:opacity-0 sm:group-hover:opacity-100 sm:focus-within:opacity-100` — always visible on mobile/touch devices, hover-reveal only on desktop. Keyboard focus within the action buttons also reveals them.

### [P2] Restore non-hover access to department actions
Same responsive opacity fix applied to the department table action buttons.

### [P3] Keep the department editor modal keyboard-dismissible
Added `Escape` key handler via `useEffect`, `useRef` with auto-focus on mount, and `tabIndex={-1}` on the dialog container. Also applied the same Escape handler to the delete confirmation modal.

### [P3] Align staff column label with rendered data
Fixed column header from "Role / Dept" to "Position / Dept" to match the actual cell content (`user.position` / `user.department`).

chrisgen19:
Current review findings on `a73d464`:

- [P2] Staff row actions are still hidden on touch-only viewports at `sm` and up — `app/(dashboard)/dashboard/users/_components/user-list-client.tsx:179`
  The action group still uses `sm:opacity-0 sm:group-hover:opacity-100`. That only makes the controls always visible below `sm`; on larger touch devices there is still no hover state to reveal the only row actions, so admins can still be blocked from view/edit/activate/deactivate flows without a mouse.

- [P2] Department row actions have the same touch-device regression — `app/(dashboard)/dashboard/departments/_components/department-table.tsx:101`
  The department action group still uses the same hover-only pattern. `focus-within` helps keyboard users once a control is focused, but on large touch-only devices the edit/delete controls can still remain inaccessible because there is no hover state to expose them.

- [P3] Department editor still is not a proper modal — `app/(dashboard)/dashboard/departments/_components/department-form.tsx:90`
  `Escape` handling was restored, but this is still a plain fixed `<div>` rather than the dialog primitive used elsewhere. Focus is not trapped inside the editor, so keyboard users can still tab behind the overlay. That leaves the modal-behavior regression unresolved.

The prior column-label mismatch appears fixed: the header now says `Position / Dept`, which matches the cell content.

chrisgen19:
## Fixes Applied (e0ec718)

All 3 review issues addressed:

### [P2] Staff & department row actions hidden on touch-only devices
Replaced `sm:opacity-0 sm:group-hover:opacity-100` with `[@media(hover:hover)]:opacity-0 [@media(hover:hover)]:group-hover:opacity-100`. This uses the CSS `hover` media feature instead of a screen-size breakpoint — actions are now always visible on touch-only devices (including large tablets) and only hover-reveal on devices with a pointer that supports hover.

### [P3] Department dialogs not proper modals
Replaced both the department editor and delete confirmation plain `<div>` modals with the shadcn `Dialog`/`DialogContent` component (base-ui primitive). This restores:
- **Focus trapping** — keyboard focus stays within the modal
- **Escape key dismissal** — built-in from the dialog primitive
- **Backdrop click** — handled by the overlay

M3 styling (rounded-[2rem], shadow-2xl, custom footer) applied via `className` overrides on `DialogContent`.

PR #2, opened 2026-03-28
feat(recognition): add Recognition Card feature

chrisgen19:
## Summary
- Add `RecognitionCard` Prisma model with sender/recipient relations and 5 company value booleans
- Add Recognition Card navigation item in sidebar (below Overview)
- Add `/dashboard/recognition` page with card feed (React Query) and top-right "Send Recognition Card" button
- Add `/dashboard/recognition/create` page with send form — searchable recipient combobox, message, date, and company value toggle pills
- Add Zod validation schema, server actions, and API routes (`/api/recognition`, `/api/recognition/users`)

## Test plan
- [ ] Navigate to `/dashboard/recognition` — page loads with feed (or empty state)
- [ ] Click "Send Recognition Card" — navigates to `/dashboard/recognition/create`
- [ ] Submit form without selecting a value — validation error shown
- [ ] Submit form with all fields filled — card appears in feed, toast shown, redirected back
- [ ] Verify you cannot select yourself as recipient
- [ ] Verify sidebar nav order: Overview → Recognition Card → Staff → Departments → My Profile
- [ ] Test dark mode and mobile responsive layout

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,36 +1,138 @@
-This is a [Next.js](https://nextjs.org) project bootstrapped with [`create-next-app`](https://nextjs.org/docs/app/api-reference/cli/create-next-app).
+# Access Recognition
+
+Internal employee recognition app for Access Group. Team members publicly recognize colleagues via digital thank-you cards tied to company values.
+
+## Tech Stack
+
+| Category | Technology |
+|----------|-----------|
+| Framework | Next.js 16 (App Router, Turbopack) |
+| Language | TypeScript 5 |
+| Runtime | Node.js 20+, Bun (package manager) |
+| Auth | better-auth (email/password + Google OAuth) |
+| Database | PostgreSQL + Prisma 7 |
+| Styling | Tailwind CSS v4 + shadcn/ui + Lucide React |
+| Forms | React Hook Form + Zod |
+| Client State | Zustand |
+| Server State | TanStack React Query |
+| Toasts | Sonner |
+| Linting | Biome |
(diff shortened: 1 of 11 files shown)


chrisgen19:
Review findings:

- [P1] Reuse the active-user check in the new recognition APIs — `app/api/recognition/route.ts:6`
  These handlers only verify that a session exists. Because `proxy.ts` only guards `/dashboard/*`, a deactivated user with a still-valid session cookie can call `/api/recognition` directly (and `/api/recognition/users`, which uses the same guard) and continue reading the recognition feed/staff list after their account has been disabled. `requireSession()` already enforces `isActive`, so the API routes need the same check to avoid bypassing deactivation.

- [P2] Derive the default date from the local timezone — `app/(dashboard)/dashboard/recognition/_components/recognition-form.tsx:61`
  `toISOString()` normalizes to UTC, so the prefilled value is wrong for users outside UTC around the day boundary. For example, in UTC+8/UTC+10 the form shows yesterday during the first part of the morning, and submitting without noticing stores the wrong recognition date.

- [P2] Format recognition dates without a timezone conversion — `app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx:46`
  This reparses the saved `date` into the viewer's local timezone before formatting it. When the API returns the usual JSON form of a Prisma `Date` like `2026-03-28T00:00:00.000Z`, browsers west of UTC render `Mar 27`, so the feed can show a different calendar day than the one that was stored.

PR #4, opened 2026-03-28
feat(dashboard): 2-column widget layout, stats card, and recognition inbox

chrisgen19:
## Summary
- Redesign `/dashboard` with 2-column widget layout: top row (welcome + send button), bottom row (stats card + scrollable public recognition feed)
- Add stats widget showing cards sent/received, monthly total, and top 3 most recognized people
- Add `/api/recognition/stats` endpoint with Prisma `groupBy` for leaderboard
- Restructure `/dashboard/recognition` as a personal inbox with Received/Sent tabs
- Add `?filter=received|sent` query param support to `/api/recognition`
- Refactor `RecognitionFeed` to accept `filter`, `showTitle`, `emptyTitle`, `emptyDescription` props
- "Send Recognition Card" button appears on both dashboard and recognition pages

## Test plan
- [ ] `/dashboard` — 2-column layout with welcome + send button on top, stats card left, public feed right
- [ ] Stats card shows correct sent/received counts, monthly total, and top 3 leaderboard
- [ ] Public feed shows all recognition cards in a scrollable container
- [ ] `/dashboard/recognition` — defaults to "Received" tab showing only cards sent to current user
- [ ] Switch to "Sent" tab — shows only cards sent by current user
- [ ] Empty states render correctly per tab
- [ ] "Send Recognition Card" button works on both pages
- [ ] Mobile responsive — columns stack vertically
- [ ] Verify dark mode rendering

### DIFF
--- a/app/(dashboard)/dashboard/_components/recognition-feed-widget.tsx
+++ b/app/(dashboard)/dashboard/_components/recognition-feed-widget.tsx
@@ -1,15 +1,73 @@
+"use client";
+
+import { useState } from "react";
+import { Globe, Building2 } from "lucide-react";
+import { cn } from "@/lib/utils";
 import { RecognitionFeed } from "../recognition/_components/recognition-feed";
 
+const FEED_TABS = [
+	{
+		key: "all" as const,
+		label: "Public",
+		icon: Globe,
+		emptyTitle: "No recognition cards yet",
+		emptyDescription: "Be the first to recognize a colleague!",
+	},
+	{
+		key: "department" as const,
+		label: "My Department",
+		icon: Building2,
+		emptyTitle: "No department recognitions yet",
+		emptyDescription:
+			"Recognition cards involving your department will appear here.",
+	},
+];
+
 export function RecognitionFeedWidget() {
+	const [activeTab, setActiveTab] = useState<"all" | "department">("all");
+	const currentTab = FEED_TABS.find((t) => t.key === activeTab)!;
+
 	return (
 		<div className="rounded-[2rem] border border-gray-200/60 dark:border-white/10 bg-card shadow-[0_2px_20px_-4px_rgba(0,0,0,0.03)] overflow-hidden flex flex-col">
-			<div className="px-6 pt-6 pb-2">
+			<div className="px-6 pt-6 pb-4 flex flex-col gap-4 sm:flex-row sm:items-center sm:justify-between">
 				<h3 className="text-[1.25rem] font-medium text-foreground tracking-tight">
-					Public Recognition Feed
+					Recognition Feed
 				</h3>
+				<div
+					className="flex gap-1 rounded-full bg-gray-100 dark:bg-white/5 p-1 w-fit"
+					role="tablist"
+					aria-label="Recognition feed filter"
+				>
+					{FEED_TABS.map((tab) => (
+						<button
+							key={tab.key}
+							type="button"
+							role="tab"
+							aria-selected={activeTab === tab.key}
+							onClick={() => setActiveTab(tab.key)}
+							className={cn(
+								"inline-flex items-center gap-1.5 rounded-full px-4 py-1.5 text-xs font-medium transition-all duration-200",
+								activeTab === tab.key
+									? "bg-card text-foreground shadow-sm"
+									: "text-muted-foreground hover:text-foreground",
+							)}
+						>
+							<tab.icon size={14} />
+							{tab.label}
+						</button>
+					))}
+				</div>
 			</div>
-			<div className="px-6 pb-6 overflow-y-auto max-h-[calc(100vh-12rem)]">
-				<RecognitionFeed filter="all" showTitle={false} />
+			<div
+				className="px-6 pb-6 overflow-y-auto max-h-[calc(100vh-12rem)]"
+				role="tabpanel"
+			>
+				<RecognitionFeed
+					filter={activeTab}
+					showTitle={false}
+					emptyTitle={currentTab.emptyTitle}
+					emptyDescription={currentTab.emptyDescription}
+				/>
 			</div>
 		</div>
 	);
--- a/app/(dashboard)/dashboard/_components/recognition-feed-widget.tsx
+++ b/app/(dashboard)/dashboard/_components/recognition-feed-widget.tsx
@@ -1,5 +1,3 @@
-"use client";
-
 import { RecognitionFeed } from "../recognition/_components/recognition-feed";
 
 export function RecognitionFeedWidget() {
--- a/app/(dashboard)/dashboard/_components/recognition-feed-widget.tsx
+++ b/app/(dashboard)/dashboard/_components/recognition-feed-widget.tsx
@@ -1,0 +1,18 @@
+"use client";
+
+import { RecognitionFeed } from "../recognition/_components/recognition-feed";
+
+export function RecognitionFeedWidget() {
+	return (
+		<div className="rounded-[2rem] border border-gray-200/60 dark:border-white/10 bg-card shadow-[0_2px_20px_-4px_rgba(0,0,0,0.03)] overflow-hidden flex flex-col">
+			<div className="px-6 pt-6 pb-2">
+				<h3 className="text-[1.25rem] font-medium text-foreground tracking-tight">
+					Public Recognition Feed
+				</h3>
+			</div>
+			<div className="px-6 pb-6 overflow-y-auto max-h-[calc(100vh-12rem)]">
+				<RecognitionFeed filter="all" showTitle={false} />
+			</div>
+		</div>
+	);
+}
--- a/app/(dashboard)/dashboard/_components/stats-widget.tsx
+++ b/app/(dashboard)/dashboard/_components/stats-widget.tsx
@@ -69,7 +69,7 @@ function StatsWidgetSkeleton() {
 }
 
 export function StatsWidget() {
-	const { data, isPending } = useQuery<{
+	const { data, isPending, isError } = useQuery<{
 		success: boolean;
 		data: StatsData;
 	}>({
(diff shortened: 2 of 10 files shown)


chrisgen19:
Review findings:

- [P2] Refresh `recognition-stats` after a card is sent — `app/(dashboard)/dashboard/_components/stats-widget.tsx:76`
  This adds a separate React Query cache entry for dashboard stats, but the existing send-card flow still invalidates only `['recognition-cards']`. In the common case where a user visits `/dashboard`, sends a recognition, and returns within the 60s stale window, the new widget keeps showing the old sent/monthly/top-recipient numbers until a hard refresh.

- [P2] Handle stats query failures instead of rendering zeroes — `app/(dashboard)/dashboard/_components/stats-widget.tsx:72`
  If `/api/recognition/stats` returns a non-2xx response (for example during an auth expiry or DB hiccup), `useQuery` leaves `data` undefined and `isPending` becomes false, so this component falls through to `stats?.sent ?? 0` / `stats?.received ?? 0`. That makes an outage look like legitimate zero activity, which is misleading for users and hard to diagnose.

chrisgen19:
## Review Response (commit 5f10bdb)

### Fixed

| Issue | Fix |
|-------|-----|
| **P2** Stale `recognition-stats` after sending a card | `recognition-form.tsx` now invalidates both `["recognition-cards"]` and `["recognition-stats"]` query keys via `Promise.all` after a successful send |
| **P2** Stats widget shows zeroes on API failure | Added `isError` check — widget now renders "Unable to load stats. Please try again later." instead of misleading zero counts when the API returns non-2xx or `data` is undefined |

chrisgen19:
Follow-up review on current HEAD `5f10bdb`:

No findings on the two issues from the prior review comment.

- `recognition-form.tsx` now invalidates both `["recognition-cards"]` and `["recognition-stats"]` after a successful send, so the stats widget should refresh instead of remaining stale within the cache window.
- `stats-widget.tsx` now handles query failure explicitly and renders "Unable to load stats. Please try again later." instead of falling through to zero-valued stats.

I only verified this in code; I did not run a browser-level manual repro or force an API failure.

chrisgen19:
## Code Review: `feat(dashboard): 2-column widget layout, stats card, and recognition inbox`

**8 files changed** | +387 / -50

Solid feature work — the dashboard layout, stats widget, and tabbed inbox are well-structured. Here are the findings:

---

### Bugs / Correctness

- **[P3] Recognition page lost server-side auth guard** — `recognition/page.tsx` was converted from Server to Client Component, removing the `getServerSession()` + `redirect()` check. The dashboard layout and `proxy.ts` still protect the route, so not exploitable, but reduces defense-in-depth.
- **[P3] Non-null assertion on tab lookup** — `recognition/page.tsx:29`: `TABS.find((t) => t.key === activeTab)!` — safe since `activeTab` is constrained, but a `Record<string, TabConfig>` lookup would be cleaner.

### Convention Violations

- **[P2] `recognition-feed-widget.tsx` has unnecessary `"use client"`** — No hooks, event handlers, or browser APIs. Just renders `RecognitionFeed` in a wrapper. Importing a client component from a Server Component works fine in Next.js — remove the directive.
- **[P2] Recognition page fully converted to Client Component** — CLAUDE.md: "Prefer Server Components." The page only needs client interactivity for the tab switcher. Extract the tabbed section into a `_components/recognition-inbox.tsx` client component and keep the page as a Server Component.
- **[P3] Duplicated `getInitials` utility** — Same function defined in both `stats-widget.tsx` and `recognition-feed.tsx`. Extract to a shared utility in `lib/`.

### Security

- **[P2] `monthlyTotal` stat is org-wide, not user-scoped** — `stats/route.ts:64-66`: Counts all recognition cards this month across the entire org. `sent` and `received` are correctly scoped to `userId`. If this is intentional (org pulse metric), document it. If it should be user-scoped, it's a data leak.
- **[P3] Unvalidated `filter` query param** — `api/recognition/route.ts:18`: Unrecognized values fall through to "no filter" (return all cards). Safe, but more explicit to return 400 for invalid values.

### Performance

- **[P2] Recognition feed has no query limit** — `api/recognition/route.ts:25`: `findMany` with no `take` limit. As cards grow, this fetches the entire table. Add `take: 50` or pagination.
- **[P3] No `staleTime` on React Query calls** — `stats-widget.tsx` and `recognition-feed.tsx` both use default `staleTime: 0`. Dashboard stats don't change every second — `staleTime: 30_000` would reduce unnecessary refetches during tab switches.
- **[P3] Sequential query in stats endpoint** — `stats/route.ts`: 4 queries run in parallel via `Promise.all`, but the 5th query (user names for leaderboard) runs sequentially after. Low impact for 3 results, but a JOIN-based approach would be cleaner.

### Accessibility

- **[P1] Tab buttons lack ARIA roles** — `recognition/page.tsx:52-69`: Plain `<button>` elements with no `role="tab"`, `aria-selected`, or wrapping `role="tablist"`. Screen readers won't announce these as tabs or convey the selected state.
- **[P3] Leaderboard lacks semantic structure** — `stats-widget.tsx:165-188`: Rankings render as `<div>` elements. Should use `<ol>` for an ordered list.
- **[P3] Skeleton loading states lack `aria-busy`** — `stats-widget.tsx:74-98`: No `aria-busy="true"` or `aria-label` to communicate loading to screen readers.

---

**Recommended before merge:**
1. **[P1]** Add ARIA `tablist`/`tab`/`aria-selected` to the Received/Sent tabs
2. **[P2]** Remove unnecessary `"use client"` from feed widget, consider keeping recognition page as Server Component
3. **[P2]** Add `take` limit to recognition feed query
4. **[P2]** Confirm whether `monthlyTotal` is intentionally org-wide or should be user-scoped

chrisgen19:
## Review Response (commit 6c53c97)

### Fixed

| Issue | Fix |
|-------|-----|
| **P1** Tab buttons lack ARIA roles | Added `role="tablist"`, `role="tab"`, `aria-selected`, and `role="tabpanel"` to recognition inbox tabs |
| **P2** Unnecessary `"use client"` on feed widget | Removed — component is now a Server Component |
| **P2** Recognition page fully client component | Extracted tabbed section to `_components/recognition-inbox.tsx` client component, restored page as Server Component with `getServerSession()` auth guard |
| **P2** No query limit on recognition feed | Added `take: 50` to `/api/recognition` Prisma query |
| **P3** Duplicated `getInitials` | Extracted to `lib/utils.ts`, imported in both `stats-widget.tsx` and `recognition-feed.tsx` |
| **P3** No `staleTime` on React Query calls | Added `staleTime: 30_000` to both stats and feed queries |
| **P3** Leaderboard lacks semantic structure | Changed from `<div>` to `<ol>/<li>` |
| **P3** Skeleton loading lacks `aria-busy` | Added `aria-busy="true"` and `aria-label` to skeleton state |

### Intentionally skipped

| Issue | Reason |
|-------|--------|
| **P2** `monthlyTotal` org-wide | Intentional — this is an org pulse metric showing total recognition activity across the company |
| **P3** Non-null assertion on tab lookup | Safe since `activeTab` is constrained to `"received" \| "sent"` by the state type |
| **P3** Unvalidated filter query param | Unrecognized values fall through to "all cards" which is safe and the default behavior |
| **P3** Sequential stats query for user names | Only fetches 3 user records — negligible impact |

chrisgen19:
## Re-review: All claimed fixes verified ✅

Checked commit `6c53c97` against each original finding:

| Fix | Verified |
|-----|----------|
| ARIA tabs (`role="tablist"`, `role="tab"`, `aria-selected`, `role="tabpanel"`) | ✅ `recognition-inbox.tsx:34-40,45-46,60` |
| `"use client"` removed from feed widget | ✅ `recognition-feed-widget.tsx` — no directive, is now a Server Component |
| Recognition page restored as Server Component with auth guard | ✅ `recognition/page.tsx` — `async`, `getServerSession()`, `redirect()` intact |
| Tab UI extracted to `recognition-inbox.tsx` client component | ✅ New file with `"use client"`, `useState`, ARIA roles |
| `take: 50` added to recognition feed query | ✅ `api/recognition/route.ts:48` |
| `getInitials` extracted to `lib/utils.ts` | ✅ Imported in both `stats-widget.tsx` and `recognition-feed.tsx` |
| `staleTime: 30_000` on React Query calls | ✅ `stats-widget.tsx:86` |
| Leaderboard uses `<ol>/<li>` | ✅ `stats-widget.tsx:140-142` |
| Skeleton has `aria-busy` + `aria-label` | ✅ `stats-widget.tsx:50` |

**Intentionally skipped items** — justifications are reasonable:
- `monthlyTotal` org-wide — confirmed as intentional org pulse metric
- Non-null assertion — safe with constrained type
- Unvalidated filter param — falls through to safe default
- Sequential leaderboard query — 3 records, negligible

**LGTM** — ready to merge.

chrisgen19:
## Review: `feat(dashboard): add Public/My Department tabs to recognition feed widget` (8b66892)

**3 files changed** | +83 / -11

ARIA tabs correctly applied, department OR filter logic is sound. Two items to address:

### Issues

- **[P2] Department filter returns all cards if user has no department** — `route.ts:27-29`: If `user?.departmentId` is falsy (user not assigned to a department), `where` stays `undefined` and the query returns ALL cards — identical to the "Public" tab with no indication to the user. Should either show an empty state (e.g. `where = { id: "none" }` to match nothing) or hide the "My Department" tab when the user has no department.

- **[P2] Extra DB query on every department filter request** — `route.ts:22-25`: Fetches `user.departmentId` with a separate Prisma query on every `?filter=department` request. Check if `session.user.departmentId` is already available from `requireSession()` to avoid the extra query.

- **[P3] `where` typed as `object | undefined`** — `route.ts:18`: Using `object` loses type safety. Use `Prisma.RecognitionCardWhereInput` from the generated client instead.

### What's good
- ARIA tabs correctly applied — consistent with the inbox tabs pattern
- `OR` filter logic for department is correct
- `RecognitionFeed` filter type properly extended
- Empty state props passed per tab

chrisgen19:
## Review Response (commit aca2d0c)

### Fixed

| Issue | Fix |
|-------|-----|
| **P2** Department filter returns all cards if user has no department | Returns empty result (`where = { id: "none" }`) when `departmentId` is null |
| **P2** Extra DB query on every department filter | Uses `session.user.departmentId` directly from the session instead of a separate Prisma `findUnique` |
| **P3** `where` typed as `object` | Changed to `Prisma.RecognitionCardWhereInput` from the generated client |

chrisgen19:
## Re-review: Department filter fixes verified ✅ (aca2d0c)

| Fix | Verified |
|-----|----------|
| Empty result for no department | ✅ `where = { id: "none" }` when `departmentId` is falsy |
| Uses `session.user.departmentId` directly | ✅ No extra Prisma query — reads from session |
| Typed as `Prisma.RecognitionCardWhereInput` | ✅ Imported from generated client |

LGTM.

PR #6, opened 2026-03-28
feat(recognition): shareable recognition cards with public page and OG image

chrisgen19:
## Summary
- After submitting a recognition card, a **share dialog** appears with a copyable URL instead of redirecting
- Share URL leads to a **fully public page** (`/recognition/{id}`) displaying the full-size branded card (front + back) — no login required
- **Dynamic OG image** generated at `/recognition/{id}/opengraph-image` for rich previews in Teams, Slack, email, and social platforms
- OG image shows card content: recipient, message, sender, date, and company values with Access Group branding

## New files
- `app/(dashboard)/dashboard/recognition/_components/share-dialog.tsx` — Share modal with URL input + copy button
- `app/recognition/[id]/page.tsx` — Public share page with Card 1 (front) + Card 2 (back) at full size
- `app/recognition/[id]/opengraph-image.tsx` — Dynamic 1200x630 OG image with card details

## Test plan
- [ ] Submit a recognition card → share dialog appears with URL
- [ ] Click copy button → URL copied to clipboard, toast shown
- [ ] Click "Done" → navigates to `/dashboard/recognition`
- [ ] Open share URL in incognito (no login) → both cards render correctly
- [ ] Non-existent card ID → 404 page
- [ ] Paste share URL in Slack/Teams → OG preview shows card content with image
- [ ] Mobile responsive on public share page
- [ ] OG image renders at `/recognition/{id}/opengraph-image` directly

PR #5, opened 2026-03-28
feat(recognition): mini branded card view with card preferences

chrisgen19:
## Summary
- Add mini branded Card 2 component (`recognition-card-mini.tsx`) for recognition feeds — scaled-down physical card with both columns (TO/message/FROM/DATE + logos/values)
- Add card preferences to `/dashboard/profile/preferences`: Card View (Physical/Simple) and Card Size (Compact/Normal/Expanded)
- Recognition feeds on both `/dashboard` and `/dashboard/recognition` respect the card preference
- Cards on `/dashboard/recognition` inbox constrained to `max-w-3xl` for better proportions
- Preferences persisted via zustand + localStorage

## Test plan
- [ ] `/dashboard` feed shows mini branded cards by default
- [ ] `/dashboard/recognition` Sent tab shows branded cards at `max-w-3xl`
- [ ] `/dashboard/recognition` Received tab empty state is full width
- [ ] `/dashboard/profile/preferences` — toggle Card View to Simple, feeds switch to compact list view
- [ ] Toggle Card Size — cards scale between Compact/Normal/Expanded
- [ ] Preferences persist across page refreshes
- [ ] Mobile responsive — card columns stack vertically

PR #3, opened 2026-03-28
feat(recognition): redesign create page with branded card design

chrisgen19:
## Summary
- Redesign `/dashboard/recognition/create` with the Access Group branded physical recognition card layout
- 2-step flow with progress bar: **Fill Card** (Card 2/back) → **Review & Submit** (Card 1/front)
- Card 1 (front) has inline editable fields for last-minute changes before submitting
- SVG logo assets extracted to `card-assets.tsx` (AccessGroupLogo, AccessBusinessLogo, BackgroundGraphic)
- FROM auto-filled from session, DEPARTMENT auto-derived from recipient's user record
- Searchable recipient combobox styled to match card design
- Synced company value checkboxes (small on Card 2, large on Card 2 right panel)
- Updated `/api/recognition/users` to include department data

## Test plan
- [ ] Navigate to `/dashboard/recognition/create` — see step 1 (Fill Card) with progress bar
- [ ] Fill all fields and click "Review Before Submit" — advances to step 2 with Card 1 preview
- [ ] Edit fields inline on Card 1 — changes persist
- [ ] Click "Back" — returns to step 1 with data preserved
- [ ] Click "Send Recognition" — card created, redirected to feed
- [ ] Try "Review Before Submit" with empty fields — validation errors shown, stays on step 1
- [ ] Verify logos, watermark, and Roboto 900 font render correctly
- [ ] Test mobile responsive layout

PR #7, opened 2026-03-29
feat(recognition): admin features for recognition cards

chrisgen19:
## Summary
- Add **All** tab with table view on `/dashboard/recognition/all` (admin/superadmin only)
- Convert recognition inbox tabs to **route-based navigation** (`/all`, `/received`, `/sent`) so refreshing preserves the active tab
- Add **delete** action (admin/superadmin gated, with confirmation dialog) and **edit** action (disabled for non-sender) in the table view
- Make public share page (`/recognition/[id]`) accessible without login
- Force light theme on login/register pages to prevent dark mode form

## Test plan
- [ ] Log in as STAFF — verify only Received/Sent tabs are visible, no All tab
- [ ] Log in as ADMIN/SUPERADMIN — verify All tab appears with table view and defaults as active
- [ ] Refresh on each tab route (`/all`, `/received`, `/sent`) — verify it stays on the correct tab
- [ ] Test delete action on All tab — confirm dialog appears, card is removed after confirmation
- [ ] Verify edit button is clickable only for cards the current user sent, faded/disabled otherwise
- [ ] Visit `/recognition/[card-id]` while logged out — verify the card renders without redirect
- [ ] Visit `/login` with system dark mode enabled — verify the form renders in light mode


chrisgen19:
Review findings:

- [P1] Update middleware so shared cards open when signed out — `app/recognition/[id]/page.tsx:122`
  Removing the page-level session check here does not actually make `/recognition/[id]` public, because `proxy.ts` still treats every `/recognition/:path*` request except `opengraph-image` as protected. In the main sharing scenario, someone opening a shared card while logged out will still be redirected to `/login`, so this feature remains broken until the middleware is relaxed too.

- [P2] Stop truncating the admin “All” tab to 50 cards — `app/(dashboard)/dashboard/recognition/_components/recognition-table.tsx:77`
  This table reuses `/api/recognition`, but that endpoint still applies `take: 50`. Once there are more than 50 recognition cards, the new admin “All” view silently drops older rows, which means admins cannot review, edit, or delete those cards from the screen that is supposed to show all of them.

- [P2] Avoid overwriting saved theme preference on auth pages — `app/(auth)/_components/force-light.tsx:10`
  Calling `setTheme("light")` here persists `light` into `next-themes` storage, so merely visiting `/login` or `/register` can reset a user's saved dark/system preference. That sticks whenever the cleanup never runs, for example after a refresh or tab close, and it can also fail to restore if `theme` is still `undefined` on first mount.

chrisgen19:
Fixes addressed in cbd3bea:

**P1 — Share page still blocked by proxy**
Removed `/recognition/:path*` from `proxy.ts` matcher and the `isProtected` check. The share page (`/recognition/[id]`) is now truly public — no redirect for unauthenticated users.

**P2 — Admin "All" tab capped at 50 cards**
The `take: 50` in `/api/recognition` now only applies when a filter is set (`received`, `sent`, `department`). The unfiltered admin "All" view returns all cards.

**P3 — ForceLight overwrites saved theme preference**
Replaced `setTheme("light")` with a CSS-only approach — directly removes the `dark` class and sets `colorScheme: "light"` on the `<html>` element on mount, and cleans up on unmount. This doesn't touch `next-themes` localStorage, so the user's saved dark/system preference is preserved.

chrisgen19:
Review findings:

- [P1] Restrict unlimited recognition queries to admins — `app/api/recognition/route.ts:64`
  When `filter` is omitted, `where` stays `undefined`, so this change drops the only row cap for every authenticated session. Because this route still only calls `requireSession()`, a normal `STAFF` user can hit `/api/recognition` directly and now receive the entire recognition history, not just admins using the All tab. If the unfiltered view is meant to be admin-only, the no-filter branch needs a role check before removing `take: 50`.

- [P2] Restore the previous theme when auth pages unmount — `app/(auth)/_components/force-light.tsx:8`
  For users whose saved theme is `dark` or `system` -> `dark`, this effect removes the `dark` class and forces a light `colorScheme`, but the cleanup never restores the prior theme or asks `next-themes` to reapply it. Because login/register both navigate with client-side `router.push()`, those users can sign in or register and land on `/dashboard` with the app still stuck in light mode until they hard-refresh or toggle the theme manually.

chrisgen19:
Fixes addressed in db84a03:

**P1 — Unfiltered API open to all users**
`/api/recognition` now returns `403 Forbidden` when a non-admin calls it without a filter param. Only ADMIN/SUPERADMIN can fetch all cards (unfiltered). Staff users must always pass `?filter=received` or `?filter=sent`.

**P2 — Theme not restored after auth pages**
`ForceLight` now reads the saved theme from `localStorage` on mount and fully restores it on unmount — re-applying the `dark` class and `colorScheme` for users with `dark` or `system` (prefers-dark) preference when they navigate away from login/register via client-side routing.

chrisgen19:
Review findings:

- [P1] Don't block staff requests from the public recognition feed — `app/api/recognition/route.ts:24`
  This now rejects every unfiltered `GET /api/recognition` request from non-admins, but the dashboard's Public feed still calls this endpoint without a `filter` query string via `RecognitionFeed`. In practice, STAFF users now get a 403 on their default dashboard tab and the widget falls back to its empty state instead of showing public recognitions.

- [P2] Restore the system theme when no explicit preference is stored — `app/(auth)/_components/force-light.tsx:16`
  Because the app's `ThemeProvider` defaults to `system`, many users won't have a `theme` entry in `localStorage` until they toggle it manually. In that case this cleanup returns early, leaving `html.style.colorScheme = "light"` in place after leaving `/login` or `/register`, so dark-system users stay stuck in light mode on the next page.

chrisgen19:
Fixes addressed in 3960d5d:

**P1 — Staff blocked from public recognition feed**
Removed the 403 block on unfiltered requests. All users now get `take: 50` by default (public feed works for everyone). The admin "All" table passes `?limit=none` to get unlimited results — the API only honors this when the caller is ADMIN/SUPERADMIN.

**P2 — System theme not restored when no localStorage entry**
When no `theme` key exists in localStorage (default state for most users), cleanup now defaults to `"system"` instead of returning early. This correctly re-applies dark mode for users whose OS preference is dark.

chrisgen19:
Follow-up review on current HEAD `3960d5d`:

No findings on the two issues from the prior review comment.

- `GET /api/recognition` now keeps the default `take: 50` behavior for normal unfiltered requests, and only honors `?limit=none` when the caller is ADMIN/SUPERADMIN. The dashboard public feed still fetches `/api/recognition` without extra params, while the admin table now explicitly uses `/api/recognition?limit=none`.
- `ForceLight` cleanup now defaults missing localStorage state to `"system"` and re-applies dark mode when `prefers-color-scheme: dark` matches, so users following the system theme are no longer left stuck in light mode after auth-page navigation.

I verified this in code only; I did not run a browser-level manual repro.

chrisgen19 (commented):
## Code Review: `feat(recognition): admin features for recognition cards`

**Branch:** `feature/recognition-admin` → `main` | **+751 / -97** across 16 files

### Overview

Converts recognition inbox from client-side tab state to route-based navigation (`/all`, `/received`, `/sent`), adds an admin-only table view with delete/edit actions, makes the public share page accessible without login, and forces light theme on auth pages.

---

### Positives

- **Route-based tabs** — great improvement over client-side state; tabs survive refresh and are shareable/bookmarkable
- **Clean role gating** — admin check on both server (page redirect) and API (`hasMinRole` + `unlimitedParam`)
- **Good separation** — `RecognitionTabs` as a reusable client component with icon map pattern, `RecognitionFeedClient` as a thin wrapper
- **Public share page** — removing the auth gate from `/recognition/[id]` makes sense for a shareable link

---

### Issues & Suggestions

#### P1 — Security / Correctness

1. **Unlimited query with no pagination** (`app/api/recognition/route.ts`)
   ```ts
   const res = await fetch("/api/recognition?limit=none");
   ...(unlimited ? {} : { take: 50 })
   ```
   When `limit=none`, the query fetches **all** recognition cards with no pagination. This is fine with 50 cards but will degrade as data grows. Consider adding server-side pagination (cursor or offset) to the table view, or at minimum a reasonable upper bound (e.g., `take: 500`).

2. **No `where` clause for unfiltered admin query** (`app/api/recognition/route.ts:19-21`)
   When `!filter && isAdmin && unlimitedParam`, `where` remains `undefined` — returning every card in the system. This is intentional for the "All" tab, but confirm that the `requireSession()` at the top of the handler is sufficient. If another role somehow hits `?limit=none` without the `filter` param, `unlimited` will be `false` (because `isAdmin` is `false`), and `where` will still be `undefined` — meaning **non-admin users without a filter also get all cards** (just capped at 50). Consider adding a fallback `where` for non-admin users when no filter is provided:
   ```ts
   } else if (!isAdmin) {
     // Non-admin without filter: default to their received cards
     where = { recipientId: session.user.id };
   }
   ```

3. **`ForceLight` cleanup race condition** (`app/(auth)/_components/force-light.tsx`)
   The cleanup function restores theme on unmount, but if the user navigates quickly (auth page → dashboard), the cleanup fires and may briefly flash dark mode or conflict with `next-themes`. Consider using `next-themes`' `setTheme("light")` in the auth layout instead of raw DOM manipulation, which would avoid fighting the theme provider.

#### P2 — Code Quality

4. **Duplicate session checks across sub-pages** — `received/page.tsx`, `sent/page.tsx`, and `all/page.tsx` each independently call `getServerSession()` / `requireSession()` + redirect. The parent `layout.tsx` already does this check. Since Next.js deduplicates `fetch`, but `requireSession` may not be deduped (depends on `cache()` wrapping), consider relying on the layout's auth guard or wrapping `requireSession` with React `cache()` to avoid redundant DB calls.

5. **Mixed auth utils** — `all/page.tsx` uses `requireSession()` (throws on no session) while `received/page.tsx` and `sent/page.tsx` use `getServerSession()` (returns null). Pick one pattern for consistency.

6. **Imports order** (`recognition-table.tsx`) — The `getInitials` function is defined mid-imports between two import blocks. Move it after all imports.
   ```ts
   // Current:
   function getInitials(...) { ... }
   import { type RecognitionCard, ... } from "@/lib/recognition";
   ```

7. **`icon` prop as string** (`recognition-tabs.tsx`) — Passing icon names as strings and mapping them via `ICON_MAP` works but loses type safety. Consider passing the icon component directly from the server layout (icons are just SVGs, they serialize fine as children), or type `icon` as `keyof typeof ICON_MAP`.

#### P3 — Minor / Nit

8. **`AlertDialogAction` `variant` prop** (`recognition-table.tsx`) — Verify that your `AlertDialogAction` component accepts a `variant` prop. The shadcn alert-dialog component typically wraps a `Button`, but the generated `alert-dialog.tsx` here extends `AlertDialogPrimitive.Close` — double-check this doesn't silently drop the `variant="destructive"` prop.

9. **No error state in table** — `useQuery` provides `isError`/`error` but the table only handles `isPending` and empty states. A fetch failure will render the empty state instead of an error message.

10. **Hardcoded `"STAFF"` fallback** appears in 4+ places: `(session.user.role as Role) ?? "STAFF"`. Consider a helper like `getUserRole(session)` to centralize this.

---

### Summary

The architecture is solid — route-based tabs, proper role gating, clean component decomposition. The main concern is the **unfiltered query for non-admin users without a filter param** (P1 #2) which could leak data. The unlimited fetch without pagination (P1 #1) should be addressed before the dataset grows. Everything else is polish.

chrisgen19:
## Review Findings — Fixed in `b9fea33`

All remaining findings from the code review have been addressed:

### P1 — Security / Correctness

**#2 — Non-admin unfiltered API query** ✅ Fixed
Added `else if (!isAdmin)` fallback in `app/api/recognition/route.ts` — non-admin users without a `filter` param now default to their received cards instead of getting all cards in the system.

**#3 — ForceLight raw DOM manipulation** ✅ Fixed
Rewrote `app/(auth)/_components/force-light.tsx` to use `useTheme()` from `next-themes` instead of direct DOM manipulation (`classList`, `style.colorScheme`, `localStorage`). Saves theme on mount, restores on unmount.

### P2 — Code Quality

**#4 — Duplicate session checks** ✅ Fixed
Wrapped `getServerSession` with React `cache()` in `lib/auth-utils.ts` to deduplicate calls within a single render pass. Layout and sub-page calls now share the same cached result.

**#5 — Mixed auth utils** ✅ Fixed (prior commit)
All sub-pages now consistently use `getServerSession()`.

**#6 — Imports order** ✅ Fixed (prior commit)
`getInitials` moved after all imports.

**#7 — Icon prop type safety** ✅ Fixed
Changed `icon: string` to `icon: keyof typeof ICON_MAP` in `recognition-tabs.tsx` — resolves to `"LayoutList" | "Inbox" | "Send"`, catching typos at compile time.

### P3 — Minor

**#8 — AlertDialogAction variant prop** ✅ No fix needed
`AlertDialogAction` accepts `React.ComponentProps<typeof Button>` and spreads to `<Button>` — `variant="destructive"` works correctly at both runtime and type level.

**#9 — No error state in table** ✅ Fixed
Added `isError` destructuring from `useQuery` and an error state UI in `recognition-table.tsx`.

**#10 — Hardcoded STAFF fallback** ✅ Fixed
Added `getUserRole()` helper in `lib/permissions.ts` and replaced all raw `(session.user.role as Role) ?? "STAFF"` casts across the codebase.

PR #8, opened 2026-03-29
feat(recognition): add filter bar and CSV export to admin All table

chrisgen19:
## Summary
- Add filter bar to admin recognition All table with search by name, company values toggle (AND logic), and month/year dropdowns
- Add CSV export that respects current filters and auto-downloads with structure matching `export-sample.csv`
- Add `Branch` enum (`ISO`, `PERTH`) and optional `branch` field to User model

## Changes
- **Filter bar**: search input, value toggle pills, month/year selects, clear button, export button
- **API**: server-side filtering (search, values, date range) + `export=true` param for unpaginated filtered results with department/branch
- **CSV export**: columns match sample — Department, Team Member, Role, Date Received, Receivers/Givers Branch, Given By, Message, Values (x/blank), Quarter Received
- **Schema**: `Branch` enum + `branch` field on User

## Test plan
- [ ] Filter by name, values, month/year — verify results update correctly
- [ ] Combine multiple filters — verify AND logic
- [ ] Click Export with no filters — downloads all cards as CSV
- [ ] Click Export with filters applied — CSV only contains matching cards
- [ ] Verify CSV structure matches `export-sample.csv` headers and format
- [ ] Verify layout doesn't shift when Clear button appears/disappears

### DIFF
--- a/app/(dashboard)/dashboard/recognition/_components/recognition-table.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/recognition-table.tsx
@@ -9,8 +9,10 @@ import { useSession } from "@/lib/auth-client";
 import { getUserRole, hasMinRole } from "@/lib/permissions";
 import {
 	type RecognitionCard,
+	type ExportRecognitionCard,
 	getSelectedValues,
 	formatRecognitionDate,
+	generateRecognitionCsv,
 } from "@/lib/recognition";
 import { deleteRecognitionCardAction } from "@/lib/actions/recognition-actions";
 import {
@@ -85,6 +87,7 @@ export function RecognitionTable() {
 	const [selectedValues, setSelectedValues] = useState<string[]>([]);
 	const [selectedMonth, setSelectedMonth] = useState("");
 	const [selectedYear, setSelectedYear] = useState("");
+	const [isExporting, setIsExporting] = useState(false);
 	const [shareCardId, setShareCardId] = useState<string | null>(null);
 	const [deleteCardId, setDeleteCardId] = useState<string | null>(null);
 	const [isDeleting, setIsDeleting] = useState(false);
@@ -138,6 +141,38 @@ export function RecognitionTable() {
 		}
 
 		return { dateFrom: "", dateTo: "" };
+	}
+
+	async function exportCards() {
+		setIsExporting(true);
+		try {
+			const params = new URLSearchParams({ export: "true" });
+			if (debouncedSearch) params.set("search", debouncedSearch);
+			if (selectedValues.length > 0) params.set("values", selectedValues.join(","));
+			const { dateFrom, dateTo } = getDateRange();
+			if (dateFrom) params.set("dateFrom", dateFrom);
+			if (dateTo) params.set("dateTo", dateTo);
+
+			const res = await fetch(`/api/recognition?${params}`);
+			if (!res.ok) throw new Error("Failed to fetch recognition cards");
+
+			const json = await res.json() as { success: boolean; data: ExportRecognitionCard[] };
+			if (!json.success) throw new Error("Export failed");
+
+			const csv = generateRecognitionCsv(json.data);
+			const blob = new Blob([csv], { type: "text/csv;charset=utf-8;" });
+			const url = URL.createObjectURL(blob);
+			const link = document.createElement("a");
+			link.href = url;
+			link.download = `recognition-export-${new Date().toISOString().split("T")[0]}.csv`;
+			link.click();
+			URL.revokeObjectURL(url);
+			toast.success(`Exported ${json.data.length} recognition cards`);
+		} catch {
+			toast.error("Failed to export recognition cards");
+		} finally {
+			setIsExporting(false);
+		}
 	}
 
 	const { data, isPending, isError } = useQuery<PaginatedResponse>({
@@ -228,6 +263,8 @@ export function RecognitionTable() {
 				selectedYear={selectedYear}
 				onYearChange={setSelectedYear}
 				onClear={clearFilters}
+				onExport={exportCards}
+				isExporting={isExporting}
 			/>
 
 			{cards.length === 0 && hasActiveFilters ? (
--- a/app/(dashboard)/dashboard/recognition/_components/recognition-table.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/recognition-table.tsx
@@ -252,6 +252,7 @@ export function RecognitionTable() {
 				onMonthChange={setSelectedMonth}
 				selectedYear={selectedYear}
 				onYearChange={setSelectedYear}
+				hasActiveFilters={hasActiveFilters}
 				onClear={clearFilters}
 				onExport={exportCards}
 				isExporting={isExporting}
--- a/app/(dashboard)/dashboard/recognition/_components/recognition-table.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/recognition-table.tsx
@@ -116,11 +116,8 @@ export function RecognitionTable() {
 	}
 
 	function getDateRange() {
-		const year = selectedYear
-			? Number(selectedYear)
-			: new Date().getFullYear();
-
 		if (selectedMonth && selectedYear) {
+			const year = Number(selectedYear);
 			const month = Number(selectedMonth);
 			const from = `${year}-${String(month).padStart(2, "0")}-01`;
 			const lastDay = new Date(year, month, 0).getDate();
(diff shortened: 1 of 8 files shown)


lib/recognition.ts
```diff
@@ -58,3 +66,92 @@ export function getSelectedValues(card: RecognitionCard): string[] {
...
+];
+
+export function generateRecognitionCsv(cards: ExportRecognitionCard[]): string {
+	const rows = cards.map((card) => [
+		card.recipient.department?.name ?? "",
+		`${card.recipient.firstName} ${card.recipient.lastName}`,
+		card.recipient.position ?? "",
+		formatDateForExport(card.date),
+		card.recipient.branch ?? "",
+		card.sender.branch ?? "",
```
chrisgen19, line 143: These CSV cells will stay empty for users created or edited through this app, because `createUserAction`, `updateUserAction`, and the profile/admin forms never send or persist a `branch` value. Unless user rows are backfilled elsewhere, the new Receivers Branch and Givers Branch columns will export blanks.

app/(dashboard)/dashboard/recognition/_components/recognition-table.tsx
```diff
@@ -79,16 +82,114 @@ export function RecognitionTable() {
...
+		}
+
+		return { dateFrom: "", dateTo: "" };
+	}
+
+	async function exportCards() {
+		setIsExporting(true);
+		try {
+			const params = new URLSearchParams({ export: "true" });
+			if (debouncedSearch) params.set("search", debouncedSearch);
```
chrisgen19: When the admin types in the search box and clicks Export before the 300 ms debounce fires, this still uses the previous `debouncedSearch` value. The downloaded CSV can ignore the text currently visible in the input, so export becomes inconsistent with the filter the user just entered.

app/(dashboard)/dashboard/recognition/_components/recognition-table.tsx
```diff
@@ -79,16 +82,114 @@ export function RecognitionTable() {
...
+		setDebouncedSearch("");
+		setSelectedValues([]);
+		setSelectedMonth("");
+		setSelectedYear("");
+	}
+
+	function getDateRange() {
+		const year = selectedYear
+			? Number(selectedYear)
+			: new Date().getFullYear();
```
chrisgen19: If an admin picks a month but leaves Year blank, `getDateRange()` still falls back to `new Date().getFullYear()`. That silently excludes matching cards from prior years even though no year filter is shown as active, so the visible filter state and result set diverge.

chrisgen19:
## Review fixes (542acdf)

Addressed 2 of the 3 review comments:

### 1. Export uses stale debounced search — **Fixed**
`exportCards()` now reads the raw `search` value instead of `debouncedSearch`, so the CSV export always matches what's currently visible in the input field regardless of debounce timing.

### 2. Month-only filter silently uses current year — **Fixed**
`getDateRange()` no longer falls back to `new Date().getFullYear()` when only a month is selected. Date filtering now requires at least a year to be selected. Selecting only a month applies no date filter, keeping the visible filter state and results consistent.

### 3. Branch field never persisted — **Not an issue**
The `branch` field is already persisted in both `createUserAction` (line 47 in `user-actions.ts`) and `updateUserAction` (line 77), and the admin user form includes a branch dropdown (lines 224-237 in `user-form.tsx`). The field works end-to-end — CSV export will show branch values for users that have one set.

chrisgen19:
Re-checked the latest head (`f19f8da`). The three issues raised earlier appear fixed: branch is now persisted through the create/update/profile flows and included in export, month-only filtering no longer silently defaults to the current year, and export now uses the live search input rather than the debounced value. I did one follow-up pass and found one remaining issue: the admin edit page still does not pass `branch` into `UserForm` default values, so an existing branch will not be preselected when editing a user.

chrisgen19:
One remaining follow-up issue: the admin edit page still does not pass `branch` into `UserForm` default values. In `app/(dashboard)/dashboard/users/[id]/edit/page.tsx`, the `defaultValues` object includes first name, last name, display name, phone, position, role, department, and active state, but not `branch`. Because the form reads `watch("branch") ?? "none"`, an existing branch will not be preselected when editing a user, which can make the form appear as if the user has no branch until the admin notices and reselects it.

chrisgen19:
## Fix: branch missing from edit page defaults (413882d)

**Issue:** The admin edit page (`users/[id]/edit/page.tsx`) did not pass `branch` into `UserForm` defaultValues, so an existing branch value was never preselected when editing a user.

**Fix:** Added `branch: user.branch ?? null` to the `defaultValues` object so the form correctly shows the user's current branch on load.

chrisgen19:
Re-reviewed the latest head (`413882d`) after the branch-defaults follow-up. No remaining findings from my side on this issue: the edit page now passes `branch: user.branch ?? null` into `UserForm` defaults, which matches the form's branch select behavior and should preselect the existing branch correctly. The earlier export/filter fixes also still look correct in the current head. `bunx tsc --noEmit` passes. I did not manually exercise the UI in a browser, so the only remaining gap is browser-level verification.

chrisgen19:
## Code Review

**7 files changed** | +665 / -157 | 7 commits

### Overview

Adds a filter bar (search, value toggles, month/year dropdowns) and CSV export to the admin recognition table. Also introduces a `Branch` enum on the User model with form support in both admin and profile pages. Follow-up commits fixed stale debounced search, month-only date fallback, profile form TS error, and missing branch default in edit page.

---

### Security

1. **CSV formula injection** — `escapeCsvField` in `lib/recognition.ts` handles commas, quotes, and newlines, but does not guard against formula injection (cells starting with `=`, `+`, `-`, `@`). If a `message` field contains `=CMD(...)`, Excel may execute it. Consider prefixing such cells with a single quote or tab character.

2. **Unbounded export query** — The `isExport` path in `app/api/recognition/route.ts` returns all matching cards unpaginated. While it's behind the `isAdmin` check, consider adding a reasonable `take` limit (e.g., 10,000) as a safety net against accidental or malicious full-table dumps.

### Code Quality

3. **`hasActiveFilters` duplicated** — Computed identically in both `recognition-table.tsx:104` and `recognition-filter-bar.tsx:130`. Not a bug, but a maintenance risk if the logic diverges. Consider computing it once and passing it as a prop.

4. **Profile form `onSubmit` typing** — The fix to `Record<string, string | null>` resolves the TS error but is looser than ideal. Consider using the inferred form type from `useForm` for field-level type safety.

### Minor Nits

| Location | Note |
|---|---|
| `recognition-filter-bar.tsx:165` | URL keys derived via `shortKey.charAt(0).toLowerCase() + shortKey.slice(1)` — fragile if a new value with a different naming pattern is added. Consider using `VALUE_KEY_MAP` keys directly as the source of truth. |
| `lib/recognition.ts:108-114` | `formatDateForExport` splits on `"T"` then re-parses — works for Prisma ISO strings but fragile if the date format changes. |
| `route.ts` date filter | `new Date(\`${dateTo}T23:59:59.999Z\`)` uses end-of-day UTC — if the DB stores dates in a different timezone, edge-case records could be missed or included. Acceptable for now given consistent usage. |

### Verdict

The PR is solid and well-structured. The follow-up commits addressed the key functional issues. **Approve with minor suggestions** — the CSV formula injection fix (#1) is the most actionable item, low effort to add. The rest are non-blocking improvements.

chrisgen19:
## Review fixes applied (625b834)

All issues from the code review have been addressed:

### 1. CSV formula injection — Fixed
`escapeCsvField` now prefixes cells starting with `=`, `+`, `-`, `@`, tab, or CR with a tab character to prevent Excel formula execution.

### 2. Unbounded export query — Fixed
Export query in `app/api/recognition/route.ts` now has `take: 10_000` limit as a safety cap.

### 3. Duplicated `hasActiveFilters` — Fixed
Removed the duplicate computation from `recognition-filter-bar.tsx`. It's now computed once in `recognition-table.tsx` and passed as a `hasActiveFilters` prop.

### 4. Profile form typing — Fixed
Replaced `Record<string, string | null>` with `typeof defaultValues` for proper field-level type safety in `profile-form.tsx`.

### 5. Fragile URL key derivation — Fixed
Filter bar value toggles now look up URL keys via `VALUE_KEY_MAP` (reverse lookup) instead of string manipulation (`shortKey.charAt(0).toLowerCase() + shortKey.slice(1)`).

chrisgen19:
## Re: hydration mismatch with `useTheme` in preferences page

Valid issue. The fix has been applied in PR #9 (`bugfix/fix-dark-light-mode`) since that's where the `useTheme` code lives — commit f2395a2.

**Fix:** Added a `mounted` state that defaults to `false`. `isDark` is `false` during SSR (since `mounted` is `false`), so the server always renders the enabled/light variant. After mount, `useEffect` sets `mounted = true`, and `isDark` resolves correctly from `resolvedTheme`. This ensures server and client render the same DOM on first pass, eliminating the hydration mismatch.

PR #9, opened 2026-03-29
fix(theme): prevent background color preference from overriding dark mode

chrisgen19:
## Summary
- Fix conflict between dark mode toggle and background color preference — BgProvider was overriding `--background` CSS variable with a light color (e.g. cream) even in dark mode, causing a broken UI (light background + dark-themed text/sidebar/cards)
- Disable background color swatches in preferences page when dark mode is active, with explanatory note
- Add missing `branch` field to profile page select query and user validation schema

## Changes
- **`components/shared/bg-provider.tsx`**: Check `resolvedTheme` from `next-themes` — skip `--background` override and remove inline style when dark mode is active
- **`app/(dashboard)/dashboard/profile/preferences/page.tsx`**: Disable background color buttons with 40% opacity and "Background color is only available in light mode" note when in dark mode
- **`app/(dashboard)/dashboard/profile/page.tsx`**: Add `branch: true` to Prisma select query (fixes TS error)
- **`lib/validations/user.ts`**: Add `branch` field to `createUserSchema` (fixes TS error on edit page)

## Test plan
- [ ] Light mode: select a background color → background changes immediately
- [ ] Toggle to dark mode → background switches to dark theme, preference ignored
- [ ] Toggle back to light mode → background restores to previously selected preference
- [ ] Dark mode preferences page → background swatches disabled with note
- [ ] TypeScript: `bunx tsc --noEmit` passes with 0 errors

### DIFF
--- a/app/(dashboard)/dashboard/profile/page.tsx
+++ b/app/(dashboard)/dashboard/profile/page.tsx
@@ -15,6 +15,7 @@ export default async function ProfilePage() {
 			displayName: true,
 			phone: true,
 			position: true,
+			branch: true,
 		},
 	});
 
--- a/app/(dashboard)/dashboard/profile/preferences/page.tsx
+++ b/app/(dashboard)/dashboard/profile/preferences/page.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import { useEffect, useState } from "react";
 import { Check, CreditCard, List, Minimize2, Square, Maximize2 } from "lucide-react";
 import { useTheme } from "next-themes";
 import { toast } from "sonner";
@@ -28,8 +29,13 @@ const SIZE_ICONS: Record<CardSize, React.ComponentType<{ size?: number; classNam
 };
 
 export default function PreferencesPage() {
+	const [mounted, setMounted] = useState(false);
 	const { resolvedTheme } = useTheme();
-	const isDark = resolvedTheme === "dark";
+	const isDark = mounted && resolvedTheme === "dark";
+
+	useEffect(() => {
+		setMounted(true);
+	}, []);
 	const bgColorId = usePreferencesStore((s) => s.bgColorId);
 	const setBgColor = usePreferencesStore((s) => s.setBgColor);
 	const cardView = usePreferencesStore((s) => s.cardView);
--- a/app/(dashboard)/dashboard/profile/preferences/page.tsx
+++ b/app/(dashboard)/dashboard/profile/preferences/page.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 import { Check, CreditCard, List, Minimize2, Square, Maximize2 } from "lucide-react";
+import { useTheme } from "next-themes";
 import { toast } from "sonner";
 import {
 	usePreferencesStore,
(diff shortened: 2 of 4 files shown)


chrisgen19:
Review finding: `app/(dashboard)/dashboard/profile/preferences/page.tsx` is hydration-unsafe with `next-themes`.

`useTheme().resolvedTheme` is `undefined` during SSR, so this page always renders the light-mode variant on the server. For users whose saved/system theme resolves to dark, the client immediately re-renders different DOM here (`disabled` buttons plus the extra helper text), which triggers the hydration mismatch that `next-themes` warns about and causes a visible UI jump on every initial load of the preferences page.

This should delay theme-dependent rendering until the page is mounted instead of branching on `resolvedTheme` during the server render.

chrisgen19:
## Fix: hydration mismatch with `useTheme` in preferences page (f2395a2)

**Issue:** `useTheme().resolvedTheme` is `undefined` during SSR, so the server always rendered the light-mode variant. For dark-mode users, the client immediately re-rendered with `disabled` buttons and the helper text, triggering a hydration mismatch.

**Fix:** Added a `mounted` state that defaults to `false`. `isDark` is `false` during SSR (since `mounted` is `false`), so server and client render the same DOM on first pass. After mount, `useEffect` sets `mounted = true` and `isDark` resolves correctly from `resolvedTheme`.

chrisgen19:
Re-reviewed the latest changes. The hydration concern in `app/(dashboard)/dashboard/profile/preferences/page.tsx` now looks addressed: the new `mounted` guard prevents the server render and initial client render from branching on `useTheme().resolvedTheme`, which removes the mismatch path for dark-theme users. I did not find any new blocking issues in the current PR diff, and the branch builds successfully.

chrisgen19:
## Code Review

**4 files changed** | +21 / -2 | 2 commits

### Overview

Fixes a conflict where `BgProvider` overrode `--background` CSS variable with a light color (e.g. cream) even in dark mode, causing broken contrast. Includes a hydration-safe `mounted` guard on the preferences page and two schema fixes (`branch` in profile select and user validation).

### Correctness

- **`bg-provider.tsx`** — Clean and correct. `resolvedTheme` check in `useEffect` is the right approach since `useEffect` only runs client-side, no SSR mismatch risk. Early `return` with `removeProperty` handles cleanup correctly.
- **`preferences/page.tsx`** — Hydration fix is correct. `mounted` starts `false`, so `isDark` is always `false` during SSR. After mount, it resolves correctly. This is the standard `next-themes` pattern.
- **Schema fixes** — `branch` additions to `user.ts` validation and `profile/page.tsx` select are straightforward and prevent TS errors.

### Code Quality

- `bg-provider.tsx` cleanup path: when toggling from light to dark, both the previous effect's cleanup and the new effect's `removeProperty` fire. Harmless (double remove is a no-op) — logic is sound.
- `isDark && "opacity-40 cursor-not-allowed"` combined with `disabled` attribute — both prevent interaction and provide visual cue. Clean approach.
- Minor nit: the `mounted` `useEffect` and the store hooks have no blank line between them (cosmetic only).

### Verdict

**Approve.** Clean, minimal fix using well-established `next-themes` patterns. No issues found.

chrisgen19:
Re-reviewed the current PR diff. I do not see any remaining correctness issues from my side. The added branch field now flows through the relevant form validation and persistence paths, and the changed files type-check cleanly.

PR #11, opened 2026-03-29
feat(auth): add super admin OAuth provider toggle settings

chrisgen19:
## Summary
- Add super admin panel to enable/disable Google and Microsoft OAuth providers on the login page
- Add Microsoft OAuth support alongside existing Google sign-in
- Store OAuth toggle settings in a new `AppSetting` model, managed via server actions
- Login page dynamically fetches OAuth settings and conditionally renders provider buttons

## Changes
- **`prisma/schema.prisma`** — Add `AppSetting` model for key-value settings storage
- **`lib/actions/settings-actions.ts`** — Server actions for reading/updating OAuth settings (SUPERADMIN only)
- **`lib/auth.ts`** — Conditionally register Microsoft provider when env vars are present
- **`env.ts`** — Add optional `MICROSOFT_CLIENT_ID` and `MICROSOFT_CLIENT_SECRET`
- **`app/api/settings/oauth/route.ts`** — Public GET endpoint for login page to fetch OAuth state
- **`app/(dashboard)/dashboard/super-admin/`** — Replace placeholder with `OAuthSettingsPanel` (toggle switches per provider)
- **`app/(auth)/login/_components/login-form.tsx`** — Fetch OAuth settings on mount, conditionally render Google/Microsoft buttons, add Microsoft sign-in handler
- **`components/ui/switch.tsx`** — New shadcn/ui Switch component

## Test plan
- [ ] Super admin can toggle Google/Microsoft OAuth on and off
- [ ] Login page shows only enabled providers
- [ ] Login page shows no OAuth section when both are disabled
- [ ] Microsoft sign-in works when enabled and env vars are configured
- [ ] Non-super-admin users cannot access the toggle endpoint
- [ ] Email/password login still works regardless of OAuth settings

### DIFF
--- a/app/(dashboard)/dashboard/super-admin/_components/oauth-settings.tsx
+++ b/app/(dashboard)/dashboard/super-admin/_components/oauth-settings.tsx
@@ -48,10 +48,22 @@ const PROVIDERS = [
 	},
 ];
 
+interface ProviderAvailability {
+	google: boolean;
+	microsoft: boolean;
+}
+
+const AVAILABILITY_KEY_MAP: Record<string, keyof ProviderAvailability> = {
+	oauth_google_enabled: "google",
+	oauth_microsoft_enabled: "microsoft",
+};
+
 export function OAuthSettingsPanel({
 	initialSettings,
+	providerAvailability,
 }: {
 	initialSettings: OAuthSettings;
+	providerAvailability: ProviderAvailability;
 }) {
 	const [settings, setSettings] = useState(initialSettings);
 	const [isPending, startTransition] = useTransition();
@@ -89,7 +101,9 @@ export function OAuthSettingsPanel({
 
 			<div className="px-8 py-6 space-y-4">
 				{PROVIDERS.map((provider) => {
-					const isEnabled = settings[provider.key];
+					const availabilityKey = AVAILABILITY_KEY_MAP[provider.key];
+					const isConfigured = providerAvailability[availabilityKey];
+					const isEnabled = isConfigured && settings[provider.key];
 					const isLoading = isPending && pendingKey === provider.key;
 
 					return (
@@ -97,6 +111,7 @@ export function OAuthSettingsPanel({
 							key={provider.key}
 							className={cn(
 								"flex items-center justify-between rounded-2xl border p-5 transition-all duration-200",
+								!isConfigured && "opacity-60",
 								isEnabled
 									? "border-primary/20 bg-primary/5"
 									: "border-gray-200 dark:border-white/10",
@@ -111,7 +126,9 @@ export function OAuthSettingsPanel({
 										{provider.name}
 									</p>
 									<p className="text-xs text-muted-foreground">
-										{provider.description}
+										{isConfigured
+											? provider.description
+											: "Not configured — environment variables are missing"}
 									</p>
 								</div>
 							</div>
@@ -123,12 +140,12 @@ export function OAuthSettingsPanel({
 										isEnabled ? "text-primary" : "text-muted-foreground",
 									)}
 								>
-									{isEnabled ? "Enabled" : "Disabled"}
+									{!isConfigured ? "Unavailable" : isEnabled ? "Enabled" : "Disabled"}
 								</span>
 								<Switch
 									checked={isEnabled}
 									onCheckedChange={(checked) => handleToggle(provider.key, checked)}
-									disabled={isLoading}
+									disabled={isLoading || !isConfigured}
 								/>
 							</div>
 						</div>
--- a/app/(dashboard)/dashboard/super-admin/_components/oauth-settings.tsx
+++ b/app/(dashboard)/dashboard/super-admin/_components/oauth-settings.tsx
@@ -1,0 +1,140 @@
+"use client";
+
+import { useState, useTransition } from "react";
+import { toast } from "sonner";
+import { Switch } from "@/components/ui/switch";
+import { updateOAuthSetting } from "@/lib/actions/settings-actions";
+import type { OAuthSettings } from "@/lib/actions/settings-actions";
+import { cn } from "@/lib/utils";
+
+const PROVIDERS = [
+	{
+		key: "oauth_google_enabled" as const,
+		name: "Google",
+		description: "Allow users to sign in with their Google account",
+		icon: (
+			<svg className="h-5 w-5" viewBox="0 0 24 24">
+				<path
+					d="M22.56 12.25c0-.78-.07-1.53-.2-2.25H12v4.26h5.92a5.06 5.06 0 0 1-2.2 3.32v2.77h3.57c2.08-1.92 3.28-4.74 3.28-8.1z"
+					fill="#4285F4"
+				/>
+				<path
+					d="M12 23c2.97 0 5.46-.98 7.28-2.66l-3.57-2.77c-.98.66-2.23 1.06-3.71 1.06-2.86 0-5.29-1.93-6.16-4.53H2.18v2.84C3.99 20.53 7.7 23 12 23z"
+					fill="#34A853"
+				/>
+				<path
+					d="M5.84 14.09c-.22-.66-.35-1.36-.35-2.09s.13-1.43.35-2.09V7.07H2.18C1.43 8.55 1 10.22 1 12s.43 3.45 1.18 4.93l2.85-2.22.81-.62z"
+					fill="#FBBC05"
+				/>
+				<path
+					d="M12 5.38c1.62 0 3.06.56 4.21 1.64l3.15-3.15C17.45 2.09 14.97 1 12 1 7.7 1 3.99 3.47 2.18 7.07l3.66 2.84c.87-2.6 3.3-4.53 6.16-4.53z"
+					fill="#EA4335"
+				/>
+			</svg>
+		),
+	},
+	{
+		key: "oauth_microsoft_enabled" as const,
+		name: "Microsoft",
+		description: "Allow users to sign in with their Microsoft account",
+		icon: (
+			<svg className="h-5 w-5" viewBox="0 0 23 23">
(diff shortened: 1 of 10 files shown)


lib/actions/settings-actions.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export async function getOAuthSettings(): Promise<OAuthSettings> {
+	const settings = await prisma.appSetting.findMany({
+		where: { key: { in: [...OAUTH_KEYS] } },
+	});
+
+	const map = new Map(settings.map((s) => [s.key, s.value]));
+
+	return {
+		oauth_google_enabled: map.get("oauth_google_enabled") === "true",
```
chrisgen19: On an upgraded or fresh database with no `app_settings` rows yet, both lookups evaluate to `false` here. Because the login form trusts these values, previously working OAuth buttons disappear immediately after rollout until a super admin manually reenables them. The fallback needs to preserve existing provider availability when no setting row exists.

app/(dashboard)/dashboard/super-admin/_components/oauth-settings.tsx
```diff
@@ -0,0 +1,140 @@
...
+				/>
+				<path
+					d="M12 5.38c1.62 0 3.06.56 4.21 1.64l3.15-3.15C17.45 2.09 14.97 1 12 1 7.7 1 3.99 3.47 2.18 7.07l3.66 2.84c.87-2.6 3.3-4.53 6.16-4.53z"
+					fill="#EA4335"
+				/>
+			</svg>
+		),
+	},
+	{
+		key: "oauth_microsoft_enabled" as const,
```
chrisgen19, line 37: This Microsoft toggle is always rendered even when `MICROSOFT_CLIENT_ID`/`MICROSOFT_CLIENT_SECRET` are not configured. In that case `lib/auth.ts` never registers the provider, but this stored flag can still make the login page render a Microsoft button that only fails with provider-not-found when clicked. The UI should hide or disable this switch unless the provider is actually configured.

lib/auth.ts
```diff
@@ -3,24 +3,38 @@ import { prismaAdapter } from "better-auth/adapters/prisma";
...
-			clientId: env.GOOGLE_CLIENT_ID,
-			clientSecret: env.GOOGLE_CLIENT_SECRET,
-			mapProfileToUser: (profile) => ({
-				firstName: profile.given_name,
-				lastName: profile.family_name,
-				name: `${profile.given_name} ${profile.family_name}`,
-			}),
-		},
-	},
+	socialProviders,
```
chrisgen19, line 37: This only removes the provider from the login UI. `socialProviders` is still built once from env at startup, so Better Auth continues accepting direct auth requests for a provider even after a super admin disables it here. The server-side auth config needs to honor the stored toggle as well.

lib/auth.ts
```diff
@@ -3,24 +3,38 @@ import { prismaAdapter } from "better-auth/adapters/prisma";
...
+			name: `${profile.given_name} ${profile.family_name}`,
+		}),
+	},
+};
+
+if (env.MICROSOFT_CLIENT_ID && env.MICROSOFT_CLIENT_SECRET) {
+	socialProviders.microsoft = {
+		clientId: env.MICROSOFT_CLIENT_ID,
+		clientSecret: env.MICROSOFT_CLIENT_SECRET,
+		mapProfileToUser: (profile: { givenName: string; surname: string }) => ({
```
chrisgen19: When Microsoft OAuth is configured, Better Auth returns OIDC name claims as `given_name` and `family_name` for this provider, not `givenName` and `surname`. This mapping will produce `undefined undefined` and can leave the required `firstName`/`lastName` fields unset on first Microsoft sign-in.

chrisgen19:
## Fixes for PR Review Issues

All 4 review findings have been addressed in commit 59bf701:

### 1. Default OAuth settings on fresh DB
`getOAuthSettings()` now uses `!== "false"` instead of `=== "true"`, so missing DB rows default to **enabled**. Existing OAuth buttons stay visible after deployment until a super admin explicitly disables them.

### 2. Microsoft toggle without env vars
Added `getOAuthProviderAvailability()` that checks if `MICROSOFT_CLIENT_ID`/`MICROSOFT_CLIENT_SECRET` are configured. The admin panel now shows the Microsoft toggle as **"Unavailable"** with a disabled switch and a note ("Not configured — environment variables are missing") when env vars are absent. The login page also checks availability before rendering the button.

### 3. Server-side enforcement of DB toggle
The auth catch-all route (`app/api/auth/[...all]/route.ts`) now intercepts social auth requests (`/callback/*`, `/sign-in/social`). It checks both env var availability and the DB setting before delegating to better-auth — returns **403** if the provider is disabled or unconfigured. This prevents direct API calls from bypassing the admin toggle.

### 4. Microsoft profile field name mismatch
Fixed `mapProfileToUser` to use `given_name`/`family_name` (standard OIDC claim names) instead of `givenName`/`surname`. Confirmed from better-auth's Microsoft Entra ID provider source in `node_modules`.

chrisgen19:
Re-reviewed the current PR head (`59bf701`). I do not see any remaining correctness findings from my side. The earlier OAuth issues now look addressed: Microsoft profile mapping uses the OIDC field names, disabled or unconfigured providers are blocked server-side, missing `app_settings` rows default to preserving existing OAuth access, and the Microsoft toggle is no longer usable when the provider is unconfigured. `bunx tsc --noEmit` passes.

chrisgen19:
## Code Review: PR #11

**10 files changed** | +462 / -73 | 2 commits

---

### Overview

Adds a super admin panel to toggle Google/Microsoft OAuth providers on/off, stored in a new `AppSetting` key-value table. The login page dynamically fetches enabled providers and conditionally renders OAuth buttons. Includes server-side enforcement via an auth route interceptor and provider availability checks based on env vars.

---

### Correctness

- **Defaults are safe** — `!== "false"` in `getOAuthSettings()` means fresh DBs default to enabled. Good.
- **Microsoft profile mapping fixed** — Uses `given_name`/`family_name` (OIDC standard), confirmed from better-auth source.
- **Server-side enforcement works** — The auth catch-all intercepts `/callback/*` and `/sign-in/social` routes, checks both env availability and DB toggle before delegating.

**Notes:**

- **`/api/settings/oauth` is unauthenticated** — This endpoint exposes which OAuth providers are configured (`availability.microsoft`). This is low-risk (it only reveals whether env vars exist, not their values), but worth noting. The login page needs it pre-auth, so this is an acceptable trade-off.

- **`getOAuthSettings()` hits the DB on every login page load and every auth request.** No caching. For a small internal app this is fine, but if traffic grows, consider a short TTL cache (even in-memory with `revalidateTag`).

---

### Code Quality & Style

- **Follows project conventions** — Server actions, `"use client"` components, Tailwind styling, consistent patterns.
- **SVG icons duplicated** — Google and Microsoft SVG icons are defined in both `login-form.tsx` and `oauth-settings.tsx`. Consider extracting to a shared location if more components need them in the future. Not blocking.
- **`socialProviders` typed as `Record<string, unknown>`** — This loses type safety on the provider config objects. The original code had better-auth inferring types from `socialProviders` inline. This is a trade-off for conditional Microsoft registration. Acceptable given the constraint.
- **`getOAuthProviderAvailability` is not async** — Correctly synchronous since it only checks env vars. Good.

---

### Security

- **Server-side enforcement is the key win** — Without the auth catch-all interceptor, disabling a provider in the admin UI was purely cosmetic. Direct API calls could still initiate OAuth flows. The interceptor closes this gap.
- **SUPERADMIN-only writes** — `updateOAuthSetting` correctly calls `requireRole("SUPERADMIN")`. Good.
- **No secrets exposed** — The availability endpoint only reveals boolean presence of env vars, not values.

---

### Suggestions (non-blocking)

1. **Minor:** Consider a brief loading state on the login form while `oauthSettings` is `null` (before the fetch resolves). Currently, OAuth buttons simply don't render until the fetch completes, which could cause a brief layout shift.

2. **Minor:** The `OAUTH_KEYS.includes(key)` check in `updateOAuthSetting` is redundant — the `key` parameter is already typed as `OAuthKey`. The runtime check is fine as defense-in-depth for a server action.

3. **Consider:** Add `revalidateTag` or short cache to `getOAuthSettings()` if this endpoint sees frequent traffic, since it's called on every login page load and every auth request through the interceptor.

---

### Verdict

**Approve with minor suggestions.** The 4 original review issues are properly addressed. Defaults are safe, availability checks prevent misconfigured toggles, server-side enforcement closes the bypass gap, and the Microsoft profile mapping is correct.

chrisgen19:
## Follow-up Review: Issue Resolution Status

**3 commits** | 10 files | +489 / -73

---

### Original Issues — All 4 Fixed

| # | Issue | Status |
|---|-------|--------|
| 1 | Fresh DB defaults to `false` (OAuth buttons disappear) | **Fixed** — uses `!== "false"`, missing rows default to enabled |
| 2 | Microsoft toggle shows without env vars | **Fixed** — `getOAuthProviderAvailability()` checks env vars, admin panel shows "Unavailable" with disabled switch |
| 3 | Server-side auth doesn't honor DB toggle | **Fixed** — `checkOAuthAllowed()` interceptor returns 403 if provider is disabled or unconfigured |
| 4 | Microsoft profile field mismatch | **Fixed** — uses `given_name`/`family_name` (OIDC standard) |

### Previous Review Suggestions — Addressed

| # | Suggestion | Status |
|---|-----------|--------|
| 1 | Login form layout shift | **Fixed** — pulse skeleton while settings load |
| 2 | Redundant `OAUTH_KEYS.includes` check | Kept as defense-in-depth — acceptable |
| 3 | DB caching for `getOAuthSettings()` | **Fixed** — 30s in-memory TTL cache, invalidated on update |

---

### Verdict

All issues are resolved. No remaining blockers. **LGTM** ✅

PR #12, opened 2026-03-30
feat(notifications): add in-app notification system

chrisgen19:
## Summary
- Adds a `Notification` model (Prisma) with `CARD_RECEIVED`, `CARD_EDITED`, and `CARD_DELETED` types
- Creates notifications automatically when recognition cards are created, edited, or deleted
- Adds `NotificationBell` dropdown in the dashboard header showing recent notifications with unread count badge
- Adds `NotificationBadge` on the sidebar Recognition nav link
- Adds `GET /api/notifications` route fetched via React Query with 30s stale time
- Adds server actions: mark single read, mark all read, mark by card ID
- Auto-marks notifications as read when the recipient views a card detail page
- Includes seed data for development

## Test plan
- [ ] Create a recognition card and verify the recipient receives a notification
- [ ] Edit a recognition card and verify the recipient sees a "card edited" notification
- [ ] Delete a recognition card (as admin) and verify the recipient sees a "card deleted" notification
- [ ] Click a notification and verify it navigates to the card detail page and marks it as read
- [ ] Click "Mark all as read" and verify all notifications are marked read
- [ ] Verify the unread count badge appears on both the header bell and sidebar link
- [ ] Verify the badge disappears when all notifications are read
- [ ] Run `bunx prisma db push` and `bun prisma/seed.ts` to verify schema and seed work


lib/actions/recognition-actions.ts
```diff
@@ -112,7 +121,15 @@ export async function updateRecognitionCardAction(cardId: string, formData: unkn
 			},
 		});
 
+		await createNotification({
```
chrisgen19: If a sender edits a card and changes `recipientId`, this path only creates a new notification for the new recipient. Existing notifications linked to this card stay attached to the old recipient, but `/api/recognition` no longer returns that card for them. Clicking the stale notification now ends in a 404 and keeps their unread count inflated until they manually clear it. We should delete or reassign prior notifications when the recipient changes.

app/(dashboard)/dashboard/recognition/[id]/page.tsx
```diff
@@ -96,6 +96,15 @@ export default async function RecognitionDetailPage({
 
 	if (!isSender && !isRecipient && !isAdmin) notFound();
 
+	if (isRecipient) {
+		prisma.notification
```
Comment (chrisgen19) [inline app/(dashboard)/dashboard/recognition/[id]/page.tsx]: When a recipient opens a card from the recognition table or a deep link while the dashboard layout already has a warm `notifications` query, this server-side `updateMany` changes database state but never invalidates the React Query cache used by `NotificationBell` and `NotificationBadge`. The bell and sidebar badge keep showing the old unread count until the 30 second cache expires or the user fully refreshes.

lib/actions/recognition-actions.ts
```diff
@@ -47,7 +48,15 @@ export async function createRecognitionCardAction(formData: unknown) {
...
```
chrisgen19: The recognition card write has already committed before the notification insert runs. If `createNotification` fails here, this action returns `{ success: false }` even though the card was created, so a user retry can create duplicates. The same partial-commit pattern is repeated in the update and delete paths below. These writes should be wrapped in a transaction or notification failure should be handled separately from the primary mutation.

app/(dashboard)/dashboard/recognition/[id]/_components/mark-notifications-read.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { useEffect, useRef } from "react";
+import { useQueryClient } from "@tanstack/react-query";
+import { markNotificationsReadByCardAction } from "@/lib/actions/notification-actions";
+
+export function MarkNotificationsRead({ cardId }: { cardId: string }) {
+	const queryClient = useQueryClient();
+	const calledRef = useRef(false);
+
+	useEffect(() => {
+		if (calledRef.current) return;
```
Comment (chrisgen19) [inline app/(dashboard)/dashboard/recognition/[id]/_components/mark-notifications-read.tsx]: When a recipient navigates between `/dashboard/recognition/[id]` pages on the client, this component is reconciled with a new `cardId` but `calledRef.current` is preserved. After the first card sets it to `true`, the effect bails out for every later card, so those notifications stay unread until a full reload. This needs to re-run when `cardId` changes.

lib/actions/recognition-actions.ts
chrisgen19: Fixed — when the recipient changes, we now delete all prior notifications for the old recipient within the same transaction before creating the new notification. See commits 37d1027 and bf3fb56.

app/(dashboard)/dashboard/recognition/[id]/page.tsx
Comment (chrisgen19) [reply app/(dashboard)/dashboard/recognition/[id]/page.tsx]: Fixed — removed the fire-and-forget server-side `prisma.updateMany` and replaced it with a `MarkNotificationsRead` client component that calls the `markNotificationsReadByCardAction` server action, then invalidates the `["notifications"]` React Query cache so the bell and sidebar badge update immediately. See commit 37d1027.

lib/actions/recognition-actions.ts
chrisgen19: Fixed — create and update paths now wrap the card mutation + notification insert in `prisma.$transaction()` so both succeed or both roll back. For the delete path, the notification is best-effort with `.catch(() => {})` since the card is already gone and the delete should still report success. See commit bf3fb56.

app/(dashboard)/dashboard/recognition/[id]/_components/mark-notifications-read.tsx
Comment (chrisgen19) [reply app/(dashboard)/dashboard/recognition/[id]/_components/mark-notifications-read.tsx]: Fixed — replaced the boolean `calledRef` with `lastCardIdRef` that tracks the last processed `cardId`. The effect now re-runs when navigating between different card detail pages on the client. See commit bf3fb56.

lib/actions/recognition-actions.ts
```diff
@@ -143,7 +180,15 @@ export async function deleteRecognitionCardAction(cardId: string) {
 
 		await prisma.recognitionCard.delete({ where: { id: cardId } });
```
chrisgen19: Deleting the card here still leaves older `CARD_RECEIVED` and `CARD_EDITED` rows behind because `Notification.card` uses `onDelete: SetNull`. For any card that has already generated notifications, the bell will keep showing dead entries that no longer navigate anywhere, and unread counts can stay inflated because those rows remain unread alongside the new `CARD_DELETED` notification.

components/shared/notification-bell.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+				<div className="max-h-80 overflow-y-auto">
+					{notifications.length === 0 ? (
+						<div className="flex flex-col items-center justify-center py-10 text-muted-foreground">
+							<Bell size={28} className="mb-2 opacity-30" />
+							<p className="text-sm">No notifications yet</p>
+						</div>
+					) : (
+						notifications.map((notification) => (
+							<button
```
chrisgen19: These notification rows are rendered as plain buttons inside `DropdownMenuContent`, so Base UI never emits the close behavior it wires into `DropdownMenuItem`. When the user clicks a notification or the `View all` action, the dashboard layout stays mounted during `router.push`, which leaves the popup open over the destination page instead of dismissing it.

lib/actions/recognition-actions.ts
chrisgen19: Fixed — now deleting all existing notifications linked to the card (`deleteMany({ where: { cardId } })`) before deleting the card itself. This removes stale `CARD_RECEIVED`/`CARD_EDITED` entries so they no longer appear in the bell with null cardIds. See commit 173c6d4.

components/shared/notification-bell.tsx
chrisgen19: Fixed — replaced plain `<button>` elements with `DropdownMenuItem` for both notification rows and the "View all" action. The dropdown now auto-closes on click as expected by Base UI. See commit 173c6d4.

app/api/notifications/route.ts
```diff
@@ -0,0 +1,47 @@
...
+		return Response.json(
+			{ success: false, error: "Unauthorized" },
+			{ status: 401 },
+		);
+	}
+
+	try {
+		const userId = session.user.id;
+
+		const [notifications, unreadCount] = await Promise.all([
```
chrisgen19, line 18: This endpoint counts unread notifications across the whole table but only returns the 20 most recent rows. Once a user has more than 20 notifications, an older unread item can disappear from the dropdown while the bell still shows a non-zero unread badge, and there is no other notifications page where they can inspect that hidden item. This needs paging or an ordering strategy that keeps unread items visible before truncating.

lib/actions/recognition-actions.ts
```diff
@@ -103,16 +118,38 @@ export async function updateRecognitionCardAction(cardId: string, formData: unkn
...
+				},
+			});
+
+			if (recipientChanged) {
+				await tx.notification.deleteMany({
+					where: { cardId, userId: existingCard.recipientId },
+				});
+			}
+
+			await tx.notification.create({
```
chrisgen19, line 139: If `recipientId` changed, this is the first notification the new recipient sees for the card, but this path still emits `CARD_EDITED`. That user never received the original card, so they get the wrong icon and message for what is effectively a new delivery. The recipient-change path should emit a received/new-card notification instead.

app/api/notifications/route.ts
chrisgen19: Fixed — changed the query ordering to `[{ isRead: "asc" }, { createdAt: "desc" }]` so unread notifications are always surfaced first in the 20-item limit, with remaining slots filled by recent read ones. See commit 3549b7b.

lib/actions/recognition-actions.ts
chrisgen19: Fixed — when the recipient changes, the notification now emits `CARD_RECEIVED` with a "sent you a recognition card" message instead of `CARD_EDITED`. The new recipient sees the correct icon and message for what is effectively a new delivery. See commit 3549b7b.

chrisgen19 (commented):
## Code Review — `feat(notifications): add in-app notification system`

**476 additions, 16 deletions across 12 files, 5 commits**

---

### Overview

Adds an in-app notification system for the recognition card feature. Notifications are created when cards are created/edited/deleted, displayed via a header bell dropdown and sidebar badge, and auto-marked read when viewing card details. Uses React Query for client-side fetching with 30s stale time.

---

### Issues Found

**1. Delete action is not transactional** (`recognition-actions.ts:181-190`)
The `deleteMany` (notifications) and `delete` (card) are separate calls. If the card delete fails after notifications are already deleted, you've lost notification records with no card deletion. These should be wrapped in a `$transaction`.

**2. `handleNotificationClick` blocks navigation on `await`** (`notification-bell.tsx:78-84`)
When clicking an unread notification, the user waits for `markNotificationReadAction` to complete before `router.push` fires. This adds latency to every unread notification click. Consider firing the mark-read optimistically or in parallel with navigation.

**3. Duplicate `queryFn` in `NotificationBell` and `NotificationBadge`**
Both components define identical `queryFn` and `queryKey` inline. If the endpoint or shape changes, two files need updating. Consider extracting a shared `useNotifications` hook.

**4. `markNotificationsReadByCardAction` missing `revalidatePath`** (`notification-actions.ts:49-67`)
The other two mark-read actions call `revalidatePath("/dashboard")`, but `markNotificationsReadByCardAction` does not. This is inconsistent — though the client component handles cache invalidation via React Query, any Server Component rendering notifications on `/dashboard` won't reflect the change.

**5. Seed doesn't clean up existing notifications** (`seed.ts:323-348`)
The seed creates notifications for all cards on every run but never deletes existing ones first. Re-running the seed produces duplicate notification rows.

---

### Minor Observations

- **`createNotification` helper is only used in the delete path now** — create and update inline `tx.notification.create` directly. The helper could be removed to reduce indirection, or all three paths could use it consistently.
- **No error toast on `MarkNotificationsRead` failure** — the `.then()` silently succeeds or fails. Fine for a background operation, but worth noting.
- **`Notification.id` uses `uuid()` while the rest of the schema uses `cuid()`** — minor inconsistency in ID strategy.

---

### Security

- All server actions properly authenticate via `requireSession()`.
- `updateMany` with `userId` scoping prevents cross-user notification manipulation.
- Route handler returns 401 for unauthorized access.
- No raw SQL or unsanitized input concerns.

---

### Test Coverage

- No tests included. The PR description has a manual test plan, but given the complexity of the transaction logic and edge cases (recipient changes, stale notifications), unit tests for the server actions would be valuable.

---

### Verdict

The core feature is well-structured and the iterative fixes addressed real issues. The remaining items — **non-transactional delete**, **blocking navigation on click**, and **seed duplication** — are the most actionable. The duplicate query function and inconsistent `revalidatePath` are cleanup items.

chrisgen19:
## Fixes for Review Issues — commit 63b139d

All 5 issues from the code review have been addressed:

### 1. Delete action is now transactional (`recognition-actions.ts`)
Wrapped `notification.deleteMany`, `recognitionCard.delete`, and `notification.create` (CARD_DELETED) in a single `prisma.$transaction()`. If any step fails, everything rolls back.

### 2. Non-blocking navigation on notification click (`notification-bell.tsx`)
`markNotificationReadAction` now fires in parallel with `router.push` instead of blocking with `await`. Navigation is instant; cache invalidation happens in the background.

### 3. Extracted shared `useNotifications` hook
Created `hooks/use-notifications.ts` with the shared `queryKey`, `queryFn`, and `staleTime` config. Both `NotificationBell` and `NotificationBadge` now use this hook — single source of truth if the endpoint or shape changes.

### 4. Added missing `revalidatePath` to `markNotificationsReadByCardAction` (`notification-actions.ts`)
Now consistent with the other two mark-read actions — calls `revalidatePath("/dashboard")` so Server Components reflect the change.

### 5. Seed cleans up existing notifications before re-seeding (`seed.ts`)
Added `prisma.notification.deleteMany()` before creating notification seed data. Re-running the seed no longer produces duplicates.

**Bonus:** Removed the now-unused `lib/notifications.ts` helper since all three paths (create, update, delete) now inline `tx.notification.create` inside their transactions.

PR #13, opened 2026-03-30
feat(recognition): add new card indicator on received tab

chrisgen19:
## Summary
- Adds a "NEW" badge and ring highlight on unread recognition cards in the received tab
- Physical card view: "NEW" pill badge on top-left corner + primary ring border
- List card view: "NEW" pill next to actions + primary border/ring highlight
- Uses unread notification `cardId`s from the shared `useNotifications` hook
- Indicator disappears once the user views the card detail page (marking it as read)

## Test plan
- [ ] Navigate to `/dashboard/recognition/received` with unread notifications
- [ ] Verify "NEW" badge appears on physical card view for unread cards
- [ ] Verify "NEW" badge and border highlight appear on list card view
- [ ] Click into a card detail page, go back — verify the indicator is gone
- [ ] Verify no indicators show on the sent tab
- [ ] Test both light and dark mode

### DIFF
--- a/app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
@@ -100,7 +100,7 @@ export function RecognitionFeed({
 }: RecognitionFeedProps) {
 	const cardView = usePreferencesStore((s) => s.cardView);
 	const cardSize = usePreferencesStore((s) => s.cardSize);
-	const { unreadCardIds } = useUnreadCardIds();
+	const { unreadCardIds } = useUnreadCardIds(filter === "received");
 	const queryParam = filter !== "all" ? `?filter=${filter}` : "";
 
 	const { data, isPending } = useQuery<{
--- a/app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
@@ -10,7 +10,7 @@ import {
 	formatRecognitionDate,
 } from "@/lib/recognition";
 import { usePreferencesStore } from "@/stores/use-preferences-store";
-import { useNotifications } from "@/hooks/use-notifications";
+import { useUnreadCardIds } from "@/hooks/use-unread-card-ids";
 import { RecognitionCardMini } from "./recognition-card-mini";
 
 function CardSkeleton() {
@@ -100,14 +100,8 @@ export function RecognitionFeed({
 }: RecognitionFeedProps) {
 	const cardView = usePreferencesStore((s) => s.cardView);
 	const cardSize = usePreferencesStore((s) => s.cardSize);
-	const { notifications } = useNotifications();
+	const { unreadCardIds } = useUnreadCardIds();
 	const queryParam = filter !== "all" ? `?filter=${filter}` : "";
-
-	const unreadCardIds = new Set(
-		notifications
-			.filter((n) => !n.isRead && n.cardId)
-			.map((n) => n.cardId as string),
-	);
 
 	const { data, isPending } = useQuery<{
 		success: boolean;
--- a/app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
@@ -10,6 +10,7 @@ import {
 	formatRecognitionDate,
 } from "@/lib/recognition";
 import { usePreferencesStore } from "@/stores/use-preferences-store";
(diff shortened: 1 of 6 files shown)


app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
```diff
@@ -99,8 +100,15 @@ export function RecognitionFeed({
 }: RecognitionFeedProps) {
 	const cardView = usePreferencesStore((s) => s.cardView);
 	const cardSize = usePreferencesStore((s) => s.cardSize);
+	const { notifications } = useNotifications();
 	const queryParam = filter !== "all" ? `?filter=${filter}` : "";
 
+	const unreadCardIds = new Set(
```
chrisgen19: `unreadCardIds` currently includes every unread notification with a `cardId`, so unread `CARD_EDITED` notifications also mark cards as `New`. That makes edited older cards look like first-time deliveries. This badge should be derived only from unread received/new-card notifications, not from the generic notification list.

app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
```diff
@@ -99,8 +100,15 @@ export function RecognitionFeed({
 }: RecognitionFeedProps) {
 	const cardView = usePreferencesStore((s) => s.cardView);
 	const cardSize = usePreferencesStore((s) => s.cardSize);
+	const { notifications } = useNotifications();
```
chrisgen19: This `New` badge is derived from `useNotifications()`, but `/api/notifications` only returns the latest 20 rows while the received feed can render up to 50 cards. Once a user has more than 20 unread card notifications, older unread received cards stop showing the badge even though they are still unread. This needs paging or a dedicated unread-card source instead of the truncated notification feed.

app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
chrisgen19: Fixed — replaced `useNotifications` with a dedicated `GET /api/notifications/unread-cards` endpoint that only returns card IDs from unread `CARD_RECEIVED` notifications. `CARD_EDITED` notifications no longer trigger the "New" badge. See commit 47817c8.

app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
chrisgen19: Fixed — the new `GET /api/notifications/unread-cards` endpoint has no row limit, returning all unread `CARD_RECEIVED` card IDs. This is consumed via a dedicated `useUnreadCardIds` hook, decoupled from the 20-item notification bell feed. See commit 47817c8.

app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
```diff
@@ -99,6 +100,7 @@ export function RecognitionFeed({
 }: RecognitionFeedProps) {
 	const cardView = usePreferencesStore((s) => s.cardView);
 	const cardSize = usePreferencesStore((s) => s.cardSize);
+	const { unreadCardIds } = useUnreadCardIds();
```
chrisgen19: `useUnreadCardIds()` is invoked for every `RecognitionFeed`, but the result is only used in the `filter === "received"` branches below. That means the dashboard widget (`all` / `department`) and the Sent inbox now make an extra `/api/notifications/unread-cards` request on every visit even though those views can never render a `New` badge. This should be gated to the received feed so the extra database-backed query is not paid on common non-received pages.

chrisgen19: Fixed — `useUnreadCardIds` now accepts an `enabled` parameter, and the feed passes `filter === "received"`. The query is skipped entirely on all/sent/department views, returning a static empty set with no API call. See commit 7afca03.

chrisgen19:
## Fixes Summary

### Commit 47817c8 — Dedicated unread-cards endpoint
- Added `GET /api/notifications/unread-cards` that returns only `CARD_RECEIVED` unread card IDs with no row limit
- Created `useUnreadCardIds` hook, replacing the `useNotifications` dependency in the feed
- `CARD_EDITED` notifications no longer trigger the "New" badge
- No 20-item truncation — all unread received card IDs are returned
- Added `unread-card-ids` cache invalidation in `MarkNotificationsRead` and `NotificationBell`

### Commit 7afca03 — Gate hook to received tab only
- `useUnreadCardIds` now accepts an `enabled` parameter
- Feed passes `filter === "received"` — the query is skipped entirely on all/sent/department views
- Returns a static empty set with no API call on non-received pages

PR #14, opened 2026-03-30
feat(auth): add OAuth account linking for Google and Microsoft

chrisgen19:
## Summary
- Enable users to link/unlink Google and Microsoft OAuth accounts from **Profile → Connected Accounts**
- Added `accountLinking` config to better-auth server setup
- Exported `linkSocial`, `listAccounts`, `unlinkAccount` from auth client
- Fixed missing `await` on `getOAuthProviderAvailability()` in both `/api/auth/[...all]` and `/api/settings/oauth` route handlers (was causing `availability` to serialize as `{}`)

## Test plan
- [ ] Navigate to Profile → Connected Accounts
- [ ] Verify email/password shows as "Active" if the user has a credential account
- [ ] Click "Connect" on Microsoft — should redirect to Microsoft OAuth flow and back
- [ ] Verify Microsoft shows as connected after linking
- [ ] Click "Disconnect" on Microsoft — should unlink and show "Connect" again
- [ ] Verify you cannot disconnect your last login method
- [ ] Verify providers not configured or disabled by admin show "Not available"

### DIFF
--- a/app/(dashboard)/dashboard/profile/connected-accounts/_components/connected-accounts-panel.tsx
+++ b/app/(dashboard)/dashboard/profile/connected-accounts/_components/connected-accounts-panel.tsx
@@ -63,10 +63,15 @@ export function ConnectedAccountsPanel({ providers, hasPassword }: ConnectedAcco
 	async function handleLink(providerId: string) {
 		setLoadingProvider(providerId);
 		try {
-			await linkSocial({
+			const result = await linkSocial({
 				provider: providerId as "google" | "microsoft",
 				callbackURL: "/dashboard/profile/connected-accounts",
+				errorCallbackURL: "/dashboard/profile/connected-accounts",
 			});
+			if (result.error) {
+				toast.error(result.error.message ?? `Failed to link ${providerId} account`);
+				setLoadingProvider(null);
+			}
 		} catch {
 			toast.error(`Failed to link ${providerId} account`);
 			setLoadingProvider(null);
--- a/app/(dashboard)/dashboard/profile/connected-accounts/_components/connected-accounts-panel.tsx
+++ b/app/(dashboard)/dashboard/profile/connected-accounts/_components/connected-accounts-panel.tsx
@@ -70,10 +70,13 @@ export function ConnectedAccountsPanel({ providers, hasPassword }: ConnectedAcco
 			});
 			if (result.error) {
 				toast.error(result.error.message ?? `Failed to link ${providerId} account`);
-				setLoadingProvider(null);
+			} else {
+				toast.success(`${providerId} account linked`);
+				router.refresh();
 			}
 		} catch {
 			toast.error(`Failed to link ${providerId} account`);
+		} finally {
 			setLoadingProvider(null);
 		}
 	}
--- a/app/(dashboard)/dashboard/profile/connected-accounts/_components/connected-accounts-panel.tsx
+++ b/app/(dashboard)/dashboard/profile/connected-accounts/_components/connected-accounts-panel.tsx
@@ -1,0 +1,185 @@
+"use client";
+
+import { useState } from "react";
+import { useRouter } from "next/navigation";
+import { toast } from "sonner";
+import { Loader2, Link2, Unlink } from "lucide-react";
+import { linkSocial, unlinkAccount } from "@/lib/auth-client";
+
+const GOOGLE_ICON = (
+	<svg className="h-5 w-5" viewBox="0 0 24 24">
+		<path
+			d="M22.56 12.25c0-.78-.07-1.53-.2-2.25H12v4.26h5.92a5.06 5.06 0 0 1-2.2 3.32v2.77h3.57c2.08-1.92 3.28-4.74 3.28-8.1z"
+			fill="#4285F4"
+		/>
+		<path
+			d="M12 23c2.97 0 5.46-.98 7.28-2.66l-3.57-2.77c-.98.66-2.23 1.06-3.71 1.06-2.86 0-5.29-1.93-6.16-4.53H2.18v2.84C3.99 20.53 7.7 23 12 23z"
+			fill="#34A853"
+		/>
+		<path
+			d="M5.84 14.09c-.22-.66-.35-1.36-.35-2.09s.13-1.43.35-2.09V7.07H2.18C1.43 8.55 1 10.22 1 12s.43 3.45 1.18 4.93l2.85-2.22.81-.62z"
+			fill="#FBBC05"
+		/>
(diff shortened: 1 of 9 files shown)


app/(dashboard)/dashboard/profile/connected-accounts/_components/connected-accounts-panel.tsx
```diff
@@ -0,0 +1,185 @@
...
+	const router = useRouter();
+	const [loadingProvider, setLoadingProvider] = useState<string | null>(null);
+
+	const linkedCount = providers.filter((p) => p.linked).length;
+	const totalLinked = linkedCount + (hasPassword ? 1 : 0);
+
+	async function handleLink(providerId: string) {
+		setLoadingProvider(providerId);
+		try {
+			await linkSocial({
```
chrisgen19: `linkSocial` can resolve with an error payload or another non-redirect response instead of throwing. In those cases this handler never clears `loadingProvider` or refreshes the page, so every Connect button stays disabled and spinning until the user reloads manually. It should inspect the returned result and reset local state on non-redirect outcomes.

app/(dashboard)/dashboard/profile/connected-accounts/_components/connected-accounts-panel.tsx
```diff
@@ -0,0 +1,185 @@
...
+				provider: providerId as "google" | "microsoft",
+				callbackURL: "/dashboard/profile/connected-accounts",
```
chrisgen19, line 39: If the OAuth round-trip fails after leaving this page, Better Auth redirects to `errorCallbackURL` and falls back to `/error` when it is omitted. This app does not define an `/error` page, so link failures here send users to a 404 instead of back to Connected Accounts where they can recover. Please pass an explicit error callback back to this page.

app/(dashboard)/dashboard/profile/connected-accounts/_components/connected-accounts-panel.tsx
chrisgen19: Fixed in a5d99b7 — `handleLink` now inspects `result.error` from `linkSocial` and clears loading state + shows a toast on non-redirect error responses.

app/(dashboard)/dashboard/profile/connected-accounts/_components/connected-accounts-panel.tsx
chrisgen19: Fixed in a5d99b7 — added `errorCallbackURL: "/dashboard/profile/connected-accounts"` so failed OAuth flows redirect back to this page instead of a 404.

app/(dashboard)/dashboard/profile/connected-accounts/_components/connected-accounts-panel.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+	async function handleLink(providerId: string) {
+		setLoadingProvider(providerId);
+		try {
+			const result = await linkSocial({
+				provider: providerId as "google" | "microsoft",
+				callbackURL: "/dashboard/profile/connected-accounts",
+				errorCallbackURL: "/dashboard/profile/connected-accounts",
+			});
+			if (result.error) {
```
chrisgen19, line 42: Handling `result.error` fixes API failures, but this still leaves `loadingProvider` set for successful non-redirect responses. Better Auth can return `redirect: false` on link attempts that do not leave the page, so this path should also clear local loading state and refresh if needed instead of only branching on `result.error`.

chrisgen19:
**Re: [review comment](https://github.com/chrisgen19/access-group-staff/pull/14#discussion_r3009892745)**

Fixed in d3a3d45 — moved `setLoadingProvider(null)` into a `finally` block so it always clears regardless of outcome. Successful non-redirect responses now show a toast and call `router.refresh()`.

chrisgen19:
## Code Review

**Branch:** `feature/account-linking` → `main` | **+248 / -5** across 7 files | 3 commits

---

### Overview

Adds a "Connected Accounts" page under the user profile where users can link/unlink Google and Microsoft OAuth accounts. Also fixes two pre-existing bugs where `getOAuthProviderAvailability()` was not awaited, causing `availability` to serialize as `{}`.

---

### Correctness

- **`await` fixes are valid and important** — both `route.ts` files had the same missing `await` bug on an async server action. Good catch.
- **`link-social` added to the OAuth guard regex** — correct, without this `checkOAuthAllowed` would not validate link requests against provider availability/settings.
- **`handleLink` error handling** — properly uses `finally` to always clear loading state, checks `result.error`, and includes `errorCallbackURL`. Solid.
- **`handleUnlink` client-side guard** (`totalLinked <= 1`) — good UX safeguard, though better-auth also enforces this server-side by default (won't unlink the last account). Belt-and-suspenders is fine here.

### Code Quality & Style

- Follows project conventions: tab indentation, `PascalCase` components, `kebab-case` files, co-located `_components/` pattern.
- Component is 193 lines — within the 150-line guideline if you exclude the SVG icon constants (which are inert data). Acceptable.
- SVG icons are duplicated from `login-form.tsx`. Not a blocker but worth noting.

### Suggestions

1. **Duplicate SVG icons** — `GOOGLE_ICON` and `MICROSOFT_ICON` are identical to those in `login-form.tsx`. Consider extracting them to a shared location (e.g. `components/shared/oauth-icons.tsx`) to avoid drift if they're ever updated.

2. **Toast shows raw provider ID** — In `handleLink` success path:
   ```ts
   toast.success(`${providerId} account linked`);
   ```
   This shows "google account linked" / "microsoft account linked" (lowercase). Should use the display name instead.

3. **`listAccounts` exported but unused** — `listAccounts` is exported from `auth-client.ts` but never used in this PR. Not harmful, but if it's for future use, it's technically dead code right now.

4. **`accountId` selected but unused** — In the server component:
   ```ts
   select: { providerId: true, accountId: true },
   ```
   `accountId` is never read. Only `providerId` is used. Can trim the select.

### Security

- OAuth guard properly extended to cover `link-social` route — prevents linking disabled/unconfigured providers.
- Session check via `getServerSession()` + redirect on the server component — good.
- No secrets or sensitive data exposed to the client.

### Performance

- `Promise.all` for parallel data fetching in the server component — good.
- No unnecessary client-side fetching; provider state is passed as props from the server.

### Test Coverage

- No tests included. At minimum, a unit test for the server component's provider-mapping logic would add confidence.

---

### Verdict

**Approve with minor suggestions.** The core logic is correct, bug fixes are valid, error handling is thorough, and it follows project conventions. The suggestions above are non-blocking improvements.

PR #15, opened 2026-03-30
feat(auth): add Google and Microsoft OAuth to register page

chrisgen19:
## Summary
- Added Google and Microsoft OAuth buttons to the register page, mirroring the existing login page pattern
- Uses `signIn.social()` which auto-creates new accounts via better-auth's `mapProfileToUser` (captures first name, last name from OAuth profile)
- OAuth buttons respect admin settings (enabled/disabled per provider) and show a loading skeleton while settings are fetched

## Test plan
- [ ] Visit `/register` — verify Google and Microsoft OAuth buttons appear (if enabled in admin settings)
- [ ] Disable a provider in super admin OAuth settings — verify the button disappears on register page
- [ ] Click "Continue with Microsoft" as a new user — verify account is created with correct first/last name
- [ ] Click "Continue with Google" as a new user — verify account is created with correct first/last name
- [ ] Verify loading skeleton appears briefly while OAuth settings are fetched
- [ ] Verify "Or continue with" divider only shows when at least one OAuth provider is enabled
- [ ] Verify email/password registration still works as before

### DIFF
--- a/app/(auth)/register/_components/register-form.tsx
+++ b/app/(auth)/register/_components/register-form.tsx
@@ -47,12 +47,17 @@ export function RegisterForm() {
 	async function handleGoogleSignIn() {
 		setIsGoogleLoading(true);
 		try {
-			await signIn.social({
+			const result = await signIn.social({
 				provider: "google",
 				callbackURL: "/dashboard",
+				errorCallbackURL: "/register",
 			});
+			if (result.error) {
+				toast.error(result.error.message ?? "Failed to sign up with Google");
+			}
 		} catch {
 			toast.error("Failed to sign up with Google");
+		} finally {
 			setIsGoogleLoading(false);
 		}
 	}
@@ -60,12 +65,17 @@ export function RegisterForm() {
 	async function handleMicrosoftSignIn() {
 		setIsMicrosoftLoading(true);
 		try {
-			await signIn.social({
+			const result = await signIn.social({
 				provider: "microsoft",
 				callbackURL: "/dashboard",
+				errorCallbackURL: "/register",
 			});
+			if (result.error) {
+				toast.error(result.error.message ?? "Failed to sign up with Microsoft");
+			}
 		} catch {
 			toast.error("Failed to sign up with Microsoft");
+		} finally {
 			setIsMicrosoftLoading(false);
 		}
 	}
--- a/app/(auth)/register/_components/register-form.tsx
+++ b/app/(auth)/register/_components/register-form.tsx
@@ -203,97 +203,97 @@ export function RegisterForm() {
 						</>
 					)}
 
-				<form onSubmit={handleSubmit(onSubmit)} className="space-y-6">
-					<div className="grid grid-cols-2 gap-5">
+					<form onSubmit={handleSubmit(onSubmit)} className="space-y-6">
+						<div className="grid grid-cols-2 gap-5">
+							<div>
+								<label htmlFor="firstName" className="block text-sm font-medium text-foreground/70 ml-1 mb-1.5">
+									First Name
+								</label>
+								<input id="firstName" type="text" className={inputClass} {...register("firstName")} />
+								{errors.firstName && (
+									<p className="mt-1 text-sm text-destructive">{errors.firstName.message}</p>
+								)}
+							</div>
+							<div>
+								<label htmlFor="lastName" className="block text-sm font-medium text-foreground/70 ml-1 mb-1.5">
+									Last Name
+								</label>
+								<input id="lastName" type="text" className={inputClass} {...register("lastName")} />
+								{errors.lastName && (
+									<p className="mt-1 text-sm text-destructive">{errors.lastName.message}</p>
+								)}
+							</div>
+						</div>
+
 						<div>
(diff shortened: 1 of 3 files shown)


app/(auth)/register/_components/register-form.tsx
```diff
@@ -28,6 +44,32 @@ export function RegisterForm() {
 		resolver: zodResolver(registerSchema),
 	});
 
+	async function handleGoogleSignIn() {
+		setIsGoogleLoading(true);
+		try {
+			await signIn.social({
```
chrisgen19: If `/api/auth/sign-in/social` returns a 4xx/5xx JSON response instead of throwing, `signIn.social()` resolves with `error` and this handler never clears `isGoogleLoading`. The button then spins forever and the email registration form stays disabled until a reload.

app/(auth)/register/_components/register-form.tsx
```diff
@@ -28,6 +44,32 @@ export function RegisterForm() {
...
+		} catch {
+			toast.error("Failed to sign up with Google");
+			setIsGoogleLoading(false);
+		}
+	}
+
+	async function handleMicrosoftSignIn() {
+		setIsMicrosoftLoading(true);
+		try {
+			await signIn.social({
```
chrisgen19: If the OAuth callback fails after leaving `/register` (for example the user cancels consent or the provider returns `invalid_code`), better-auth falls back to `${baseURL}/error` unless `errorCallbackURL` is provided. This app does not define an `/error` route, so the new register flow drops users onto a 404 instead of sending them back to the registration screen with feedback.

app/(auth)/register/_components/register-form.tsx
chrisgen19: Fixed in 7417192 — moved loading state reset to `finally` blocks so it always clears, even when `signIn.social()` resolves with an error instead of throwing. Also added `result.error` check to surface the error message via toast. Applied the same fix to the login form as well.

app/(auth)/register/_components/register-form.tsx
chrisgen19: Fixed in 7417192 — added `errorCallbackURL: "/register"` so failed/cancelled OAuth flows redirect back to the register page instead of a 404. Also applied the same fix to the login form with `errorCallbackURL: "/login"`.

app/(auth)/register/_components/register-form.tsx
```diff
@@ -28,6 +44,42 @@ export function RegisterForm() {
 		resolver: zodResolver(registerSchema),
 	});
 
+	async function handleGoogleSignIn() {
+		setIsGoogleLoading(true);
+		try {
+			const result = await signIn.social({
+				provider: "google",
+				callbackURL: "/dashboard",
+				errorCallbackURL: "/register",
```
chrisgen19, line 67: This avoids the 404, but callback failures still return silently. `better-auth` redirects to `errorCallbackURL` with `?error=...`, and neither `/register` nor `/login` reads that query param or turns it into UI feedback, so cases like cancelled consent or `invalid_code` still land back on the auth page with no message.

chrisgen19: Fixed in 01e8fd9 — both login and register forms now read the `?error=` query param from `errorCallbackURL` redirects and display a descriptive toast (`invalid_code`, `access_denied`, `server_error`, or a generic fallback).

chrisgen19 (commented):
## Code Review

### Issues Found

**1. Indentation bug in `register-form.tsx` (lines 206, 295-297)**

The `<form>` element at line 206 is indented with 5 tabs but should be 6 (it's inside `<div className="space-y-6">`). The closing `</div>` at line 297 is also misaligned. Doesn't break functionality but makes the JSX nesting hard to read and maintain.

**2. Duplicated `OAUTH_ERROR_MESSAGES` across both files**

The same error message map is defined in both `login-form.tsx` and `register-form.tsx`, only differing by "sign-in" vs "sign-up". Acceptable for now given it's just 3 entries with intentionally different wording per page, but worth noting.

**3. Duplicated OAuth logic across login and register forms**

The OAuth settings fetch, button rendering, handlers, and loading states are copy-pasted across both forms (~80 lines). Any future OAuth provider or behavior change needs to be applied in two places. Consider extracting a shared `OAuthButtons` component in a follow-up.

**4. `showGoogle` asymmetric logic (pre-existing, not introduced by this PR)**

```ts
const showGoogle = oauthSettings?.oauth_google_enabled && oauthAvailability?.google !== false;
const showMicrosoft = oauthSettings?.oauth_microsoft_enabled && oauthAvailability?.microsoft === true;
```

Google uses `!== false` (shows when `undefined`) while Microsoft uses `=== true` (hidden when `undefined`). Inherited from the login form — now duplicated.

### What Looks Good
- Error handling is solid: `finally` blocks, `result.error` checks, `errorCallbackURL`, and `?error` query param toasts
- `<Suspense>` boundary correctly added for `useSearchParams()`
- Button styles, loading states, and settings-based visibility match the login page exactly
- Login form also got the error handling improvements

### Recommendations
1. **Fix the indentation** on the `<form>` and closing `</div>` tags in `register-form.tsx` — should fix before merge
2. **Consider extracting a shared `OAuthButtons` component** — not blocking, but worth doing before adding more providers

PR #20, opened 2026-04-15
feat(recognition): add emoji reactions and comment section to recognition cards

### LINKED ISSUE #19: feat(recognition): add emoji reactions and comment section to recognition cards
## Summary

Add interactive emoji reactions and a collapsible comment section to recognition cards in the feed — similar to Facebook/LinkedIn interactions.

### Emoji Reactions
- Fixed set of 6 reactions: 👏 ❤️ 🔥 🎉 💪 😊
- Any authenticated user can react to any card
- Clicking a reaction toggles it on/off for the current user
- Reaction bar is always visible below the card
- Shows emoji + count grouped by type

### Comments
- Any authenticated user can post a comment
- Comments are collapsed by default — shown via "View X comments" toggle
- Comment author **or** ADMIN+ can delete a comment
- Comment author can edit their own comment

## Schema Changes

New models needed in `prisma/schema.prisma`:
- `CardReaction` — cardId, userId, emoji, createdAt
- `CardComment` — cardId, userId, body, createdAt, updatedAt

## API / Actions

New Server Actions in `lib/actions/`:
- `addReactionAction(cardId, emoji)`
- `removeReactionAction(cardId, emoji)`
- `addCommentAction(cardId, body)`
- `editCommentAction(commentId, body)`
- `deleteCommentAction(commentId)`

New API route:
- `GET /api/recognition/[cardId]/interactions` — returns reactions + comments for a card

## UI Components

- `CardInteractionBar` — reaction pills + comment toggle, lives below every card in the feed
- `CommentThread` — collapsible comment list + input form
- Works on both "physical card" and "digital" feed view modes

## Acceptance Criteria
- [ ] Users can react with any of the 6 fixed emojis; toggling removes the reaction
- [ ] Reaction counts update optimistically
- [ ] Comments are collapsed by default, expandable per card
- [ ] Users can add, edit, and delete their own comments
- [ ] ADMIN+ can delete any comment
- [ ] Works in both card view modes (physical + digital)
- [ ] Works on mobile

chrisgen19:
Closes #19

## Summary
- Fixed emoji reactions (👏 ❤️ 🔥 🎉 💪 😊) on every recognition card with optimistic UI updates
- Collapsible comment thread per card — collapsed by default, expandable inline
- Comment authors + ADMIN+ can delete comments; authors can edit their own

### Schema
- `CardReaction` model — unique constraint per `(cardId, userId, emoji)`, cascade delete
- `CardComment` model — indexed by `(cardId, createdAt)`, cascade delete

### API
- `GET /api/recognition/[cardId]/interactions` — returns reactions grouped with counts + `hasReacted` flag, and full comment thread

### Server Actions (`lib/actions/interaction-actions.ts`)
- `toggleReactionAction` — upsert/remove reaction (validated against fixed emoji set)
- `addCommentAction` / `editCommentAction` / `deleteCommentAction` — author-only edit, author+ADMIN delete

### UI Components
- `CardInteractionBar` — reaction pills with counts, ghost buttons for zero-count emojis, comment count toggle; React Query with optimistic reaction updates
- `CommentThread` — chronological comment list with add/edit/delete, inline edit mode, `timeAgo` timestamps

### Wired into
- Dashboard recognition feed (both digital + physical card views)
- Received and Sent inbox pages

## Test plan
- [ ] React to a card with each of the 6 emojis — count increments, pill highlights
- [ ] Click a highlighted emoji again — reaction removed, count decrements
- [ ] Open comments, post a comment — appears immediately
- [ ] Edit own comment — inline edit with Enter to save, Esc to cancel
- [ ] Delete own comment — disappears immediately
- [ ] As ADMIN, delete another user's comment
- [ ] As STAFF, confirm delete button is not shown on others' comments
- [ ] Verify both physical card and digital card views show the interaction bar
- [ ] Verify on mobile layout

### DIFF
--- a/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
@@ -30,9 +30,9 @@ export function CardInteractionBar({
 	const [fetchEnabled, setFetchEnabled] = useState(false);
 
 	const { data } = useQuery<{ success: boolean; data: CardInteractions }>({
-		queryKey: ["card-interactions", cardId],
-		queryFn: async () => {
-			const res = await fetch(`/api/recognition/${cardId}/interactions`);
+		queryKey: ["card-interactions", cardId, currentUserId],
+		queryFn: async ({ signal }) => {
+			const res = await fetch(`/api/recognition/${cardId}/interactions`, { signal });
 			if (!res.ok) throw new Error("Failed to fetch interactions");
 			return res.json();
 		},
@@ -50,18 +50,17 @@ export function CardInteractionBar({
 			// Data not loaded yet — action handles add/remove atomically on server;
 			// invalidate after so the query picks up the new state
 			toggleReactionAction(cardId, emoji).then((result) => {
-				if (result.success) {
-					queryClient.invalidateQueries({ queryKey: ["card-interactions", cardId] });
-				} else {
+				if (!result.success) {
 					toast.error(result.error);
 				}
+				queryClient.invalidateQueries({ queryKey: ["card-interactions", cardId, currentUserId] });
 			});
 			return;
 		}
 
 		// Optimistic update when we already have the current state
 		queryClient.setQueryData<{ success: boolean; data: CardInteractions }>(
-			["card-interactions", cardId],
+			["card-interactions", cardId, currentUserId],
 			(old) => {
 				if (!old?.data) return old;
 				const reactions = old.data.reactions.map((r) => {
@@ -79,9 +78,10 @@ export function CardInteractionBar({
 
 		toggleReactionAction(cardId, emoji).then((result) => {
 			if (!result.success) {
-				queryClient.invalidateQueries({ queryKey: ["card-interactions", cardId] });
 				toast.error(result.error);
 			}
+			// Always invalidate to pick up concurrent reactions from other users
+			queryClient.invalidateQueries({ queryKey: ["card-interactions", cardId, currentUserId] });
 		});
 	}
 
@@ -91,9 +91,9 @@ export function CardInteractionBar({
 	}
 
 	async function handleCommentsChange(comments: CardComment[]) {
-		await queryClient.cancelQueries({ queryKey: ["card-interactions", cardId] });
+		await queryClient.cancelQueries({ queryKey: ["card-interactions", cardId, currentUserId] });
 		queryClient.setQueryData<{ success: boolean; data: CardInteractions }>(
-			["card-interactions", cardId],
+			["card-interactions", cardId, currentUserId],
 			(old) => {
 				// Seed cache from scratch if the lazy query hasn't fired yet
 				if (!old?.data) {
--- a/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
@@ -18,6 +18,7 @@ interface CardInteractionBarProps {
 	currentUserId: string;
 	isAdmin: boolean;
 	initialCommentCount?: number;
+	initialReactions?: { emoji: string; count: number; hasReacted: boolean }[];
 }
 
 export function CardInteractionBar({
@@ -25,6 +26,7 @@ export function CardInteractionBar({
 	currentUserId,
 	isAdmin,
 	initialCommentCount = 0,
+	initialReactions,
 }: CardInteractionBarProps) {
 	const queryClient = useQueryClient();
 	const [showComments, setShowComments] = useState(false);
@@ -138,11 +140,17 @@ export function CardInteractionBar({
 	const comments = interactions?.comments ?? [];
 	const totalComments = interactions?.totalComments ?? initialCommentCount;
 
-	const activeReactions = reactions.filter((r) => r.count > 0);
-	// Show all 6 ghost buttons when data not loaded; only zero-count ones after load
-	const ghostReactions = interactions
-		? reactions.filter((r) => r.count === 0)
-		: REACTION_EMOJIS.map((emoji) => ({ emoji, count: 0, hasReacted: false }));
+	// Use initialReactions from the feed before the lazy fetch fires
+	const displayReactions = interactions
+		? reactions
+		: initialReactions ?? [];
+
+	const activeReactions = displayReactions.filter((r) => r.count > 0);
+	// Build ghost buttons for emojis not already shown as active
+	const activeEmojis = new Set(activeReactions.map((r) => r.emoji));
+	const ghostReactions = REACTION_EMOJIS.filter((e) => !activeEmojis.has(e)).map(
+		(emoji) => ({ emoji, count: 0, hasReacted: false }),
+	);
 
 	return (
 		<div
--- a/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
@@ -26,24 +26,40 @@ export function CardInteractionBar({
 }: CardInteractionBarProps) {
 	const queryClient = useQueryClient();
 	const [showComments, setShowComments] = useState(false);
+	// Lazy: only fetch when user first hovers (desktop) or interacts (mobile)
+	const [fetchEnabled, setFetchEnabled] = useState(false);
 
-	const { data, isLoading } = useQuery<{ success: boolean; data: CardInteractions }>({
+	const { data } = useQuery<{ success: boolean; data: CardInteractions }>({
 		queryKey: ["card-interactions", cardId],
 		queryFn: async () => {
 			const res = await fetch(`/api/recognition/${cardId}/interactions`);
 			if (!res.ok) throw new Error("Failed to fetch interactions");
 			return res.json();
 		},
+		enabled: fetchEnabled,
 		staleTime: 30_000,
 	});
 
 	const interactions = data?.data;
 
 	function handleReaction(emoji: string) {
-		const current = interactions;
-		if (!current) return;
+		// Ensure the query is (or will be) active
+		setFetchEnabled(true);
 
-		// Optimistic update
+		if (!interactions) {
+			// Data not loaded yet — action handles add/remove atomically on server;
+			// invalidate after so the query picks up the new state
+			toggleReactionAction(cardId, emoji).then((result) => {
+				if (result.success) {
+					queryClient.invalidateQueries({ queryKey: ["card-interactions", cardId] });
+				} else {
+					toast.error(result.error);
+				}
+			});
+			return;
+		}
+
+		// Optimistic update when we already have the current state
 		queryClient.setQueryData<{ success: boolean; data: CardInteractions }>(
 			["card-interactions", cardId],
 			(old) => {
@@ -63,11 +79,15 @@ export function CardInteractionBar({
 
 		toggleReactionAction(cardId, emoji).then((result) => {
 			if (!result.success) {
-				// Revert on failure
 				queryClient.invalidateQueries({ queryKey: ["card-interactions", cardId] });
 				toast.error(result.error);
 			}
 		});
+	}
+
+	function handleCommentToggle() {
+		setFetchEnabled(true);
+		setShowComments((v) => !v);
 	}
 
 	function handleCommentsChange(comments: CardComment[]) {
@@ -83,31 +103,24 @@ export function CardInteractionBar({
 		);
 	}
 
-	if (isLoading) {
-		return (
-			<div className="flex gap-1.5 pt-3 mt-3 border-t border-border/50">
-				{REACTION_EMOJIS.map((emoji) => (
-					<div
-						key={emoji}
-						className="h-7 w-10 rounded-full bg-muted/60 animate-pulse"
-					/>
-				))}
-			</div>
-		);
-	}
-
 	const reactions = interactions?.reactions ?? [];
 	const comments = interactions?.comments ?? [];
 	const totalComments = interactions?.totalComments ?? 0;
 
 	const activeReactions = reactions.filter((r) => r.count > 0);
-	const inactiveReactions = reactions.filter((r) => r.count === 0);
+	// Show all 6 ghost buttons when data not loaded; only zero-count ones after load
+	const ghostReactions = interactions
+		? reactions.filter((r) => r.count === 0)
+		: REACTION_EMOJIS.map((emoji) => ({ emoji, count: 0, hasReacted: false }));
 
 	return (
-		<div className="pt-3 mt-3 border-t border-border/50">
+		<div
+			className="pt-3 mt-3 border-t border-border/50"
+			onMouseEnter={() => setFetchEnabled(true)}
+		>
 			{/* Reaction row */}
 			<div className="flex flex-wrap items-center gap-1.5">
-				{/* Active reactions first */}
+				{/* Active reactions with counts */}
 				{activeReactions.map((r) => (
 					<button
 						key={r.emoji}
@@ -126,8 +139,8 @@ export function CardInteractionBar({
 					</button>
 				))}
 
-				{/* Inactive emojis as small ghost buttons */}
-				{inactiveReactions.map((r) => (
+				{/* Ghost buttons for zero-count emojis */}
+				{ghostReactions.map((r) => (
 					<button
 						key={r.emoji}
 						type="button"
@@ -142,7 +155,7 @@ export function CardInteractionBar({
 				{/* Comment toggle */}
 				<button
 					type="button"
-					onClick={() => setShowComments((v) => !v)}
+					onClick={handleCommentToggle}
 					className={cn(
 						"ml-auto inline-flex items-center gap-1.5 rounded-full px-2.5 py-1 text-xs transition-colors",
 						showComments
--- a/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
@@ -1,0 +1,177 @@
+"use client";
+
+import { useState } from "react";
+import { MessageCircle, ChevronDown, ChevronUp } from "lucide-react";
+import { useQuery, useQueryClient } from "@tanstack/react-query";
+import { toast } from "sonner";
+import { cn } from "@/lib/utils";
+import {
+	REACTION_EMOJIS,
+	type CardInteractions,
+	type CardComment,
+} from "@/lib/recognition";
+import { toggleReactionAction } from "@/lib/actions/interaction-actions";
+import { CommentThread } from "./comment-thread";
+
+interface CardInteractionBarProps {
+	cardId: string;
+	currentUserId: string;
+	isAdmin: boolean;
+}
+
+export function CardInteractionBar({
+	cardId,
+	currentUserId,
+	isAdmin,
+}: CardInteractionBarProps) {
+	const queryClient = useQueryClient();
+	const [showComments, setShowComments] = useState(false);
+
+	const { data, isLoading } = useQuery<{ success: boolean; data: CardInteractions }>({
+		queryKey: ["card-interactions", cardId],
+		queryFn: async () => {
+			const res = await fetch(`/api/recognition/${cardId}/interactions`);
+			if (!res.ok) throw new Error("Failed to fetch interactions");
+			return res.json();
+		},
+		staleTime: 30_000,
+	});
+
+	const interactions = data?.data;
+
+	function handleReaction(emoji: string) {
+		const current = interactions;
+		if (!current) return;
+
+		// Optimistic update
+		queryClient.setQueryData<{ success: boolean; data: CardInteractions }>(
+			["card-interactions", cardId],
+			(old) => {
+				if (!old?.data) return old;
+				const reactions = old.data.reactions.map((r) => {
+					if (r.emoji !== emoji) return r;
+					const adding = !r.hasReacted;
+					return {
+						...r,
+						count: adding ? r.count + 1 : r.count - 1,
+						hasReacted: adding,
+					};
+				});
+				return { ...old, data: { ...old.data, reactions } };
+			},
+		);
+
+		toggleReactionAction(cardId, emoji).then((result) => {
+			if (!result.success) {
+				// Revert on failure
+				queryClient.invalidateQueries({ queryKey: ["card-interactions", cardId] });
+				toast.error(result.error);
+			}
+		});
+	}
+
+	function handleCommentsChange(comments: CardComment[]) {
+		queryClient.setQueryData<{ success: boolean; data: CardInteractions }>(
+			["card-interactions", cardId],
+			(old) => {
+				if (!old?.data) return old;
+				return {
+					...old,
+					data: { ...old.data, comments, totalComments: comments.length },
+				};
+			},
+		);
+	}
+
+	if (isLoading) {
+		return (
+			<div className="flex gap-1.5 pt-3 mt-3 border-t border-border/50">
+				{REACTION_EMOJIS.map((emoji) => (
+					<div
+						key={emoji}
+						className="h-7 w-10 rounded-full bg-muted/60 animate-pulse"
+					/>
+				))}
+			</div>
+		);
+	}
+
+	const reactions = interactions?.reactions ?? [];
+	const comments = interactions?.comments ?? [];
+	const totalComments = interactions?.totalComments ?? 0;
+
+	const activeReactions = reactions.filter((r) => r.count > 0);
+	const inactiveReactions = reactions.filter((r) => r.count === 0);
+
+	return (
+		<div className="pt-3 mt-3 border-t border-border/50">
+			{/* Reaction row */}
+			<div className="flex flex-wrap items-center gap-1.5">
+				{/* Active reactions first */}
+				{activeReactions.map((r) => (
+					<button
+						key={r.emoji}
+						type="button"
+						onClick={() => handleReaction(r.emoji)}
+						className={cn(
+							"inline-flex items-center gap-1 rounded-full px-2.5 py-1 text-sm transition-all",
+							r.hasReacted
+								? "bg-primary/15 ring-1 ring-primary/40 text-foreground font-medium"
+								: "bg-muted/60 dark:bg-white/5 text-foreground/80 hover:bg-muted dark:hover:bg-white/10",
+						)}
+						aria-label={`React with ${r.emoji} (${r.count})`}
+					>
+						<span>{r.emoji}</span>
+						<span className="text-xs tabular-nums">{r.count}</span>
+					</button>
+				))}
+
+				{/* Inactive emojis as small ghost buttons */}
+				{inactiveReactions.map((r) => (
+					<button
+						key={r.emoji}
+						type="button"
+						onClick={() => handleReaction(r.emoji)}
+						className="inline-flex items-center justify-center rounded-full w-7 h-7 text-sm bg-muted/40 dark:bg-white/5 hover:bg-muted dark:hover:bg-white/10 transition-colors opacity-60 hover:opacity-100"
+						aria-label={`React with ${r.emoji}`}
+					>
+						{r.emoji}
+					</button>
+				))}
+
+				{/* Comment toggle */}
+				<button
+					type="button"
+					onClick={() => setShowComments((v) => !v)}
+					className={cn(
+						"ml-auto inline-flex items-center gap-1.5 rounded-full px-2.5 py-1 text-xs transition-colors",
+						showComments
+							? "bg-muted text-foreground"
+							: "text-muted-foreground hover:text-foreground hover:bg-muted/60",
+					)}
+					aria-label="Toggle comments"
+				>
+					<MessageCircle size={13} />
+					<span>
+						{totalComments > 0
+							? `${totalComments} comment${totalComments !== 1 ? "s" : ""}`
+							: "Comment"}
+					</span>
+					{totalComments > 0 &&
+						(showComments ? <ChevronUp size={12} /> : <ChevronDown size={12} />)}
+				</button>
+			</div>
+
+			{/* Comment thread */}
+			{showComments && (
+				<CommentThread
+					cardId={cardId}
+					comments={comments}
+					currentUserId={currentUserId}
+					isAdmin={isAdmin}
+					onCommentsChange={handleCommentsChange}
+				/>
+			)}
+		</div>
+	);
+}
--- a/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
@@ -90,7 +90,8 @@ export function CardInteractionBar({
 		setShowComments((v) => !v);
 	}
 
-	function handleCommentsChange(comments: CardComment[]) {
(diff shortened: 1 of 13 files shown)


lib/actions/interaction-actions.ts
```diff
@@ -0,0 +1,179 @@
...
+		if (!parsed.success) {
+			return { success: false as const, error: "Invalid input" };
+		}
+
+		const card = await prisma.recognitionCard.findUnique({
+			where: { id: cardId },
+			select: { id: true },
+		});
+		if (!card) {
+			return { success: false as const, error: "Card not found" };
```
chrisgen19, line 36: This only verifies that the card exists here (and again in `addCommentAction`), so any authenticated staff member can post comments or reactions on any recognition card UUID they know. That bypasses the sender/recipient/admin checks already enforced elsewhere in the recognition UI; for example, a user can mutate a card from a copied `/recognition/{id}` share link even when it never appears in their inbox.

app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
```diff
@@ -0,0 +1,177 @@
...
+	const [showComments, setShowComments] = useState(false);
+
+	const { data, isLoading } = useQuery<{ success: boolean; data: CardInteractions }>({
+		queryKey: ["card-interactions", cardId],
+		queryFn: async () => {
+			const res = await fetch(`/api/recognition/${cardId}/interactions`);
+			if (!res.ok) throw new Error("Failed to fetch interactions");
+			return res.json();
+		},
+		staleTime: 30_000,
```
chrisgen19: Each `CardInteractionBar` performs its own fetch on mount, and `RecognitionFeed` now renders one bar per card. On inbox/admin pages that load up to 50 cards, a single feed render becomes 51 HTTP requests and 100 extra DB queries even when nobody opens the comment thread, which will noticeably slow the page and increase backend load. This data should be fetched lazily when the thread is expanded, or batched with the feed payload.

app/(dashboard)/dashboard/recognition/_components/comment-thread.tsx
```diff
@@ -0,0 +1,282 @@
...
+			} else {
+				toast.error(result.error);
+			}
+		});
+	}
+
+	function handleEdit(commentId: string, newBody: string) {
+		onCommentsChange(
+			comments.map((c) => (c.id === commentId ? { ...c, body: newBody } : c)),
+		);
```
chrisgen19: A successful edit only patches `body` in local state, so `updatedAt` stays equal to `createdAt` until the query refetches. As a result the `· edited` badge and relative timestamp do not update immediately after saving an edit, even though `editCommentAction` already returns the fresh timestamps.

app/api/recognition/[cardId]/interactions/route.ts
```diff
@@ -0,0 +1,90 @@
...
+				{ status: 401 },
+			);
+		}
+
+		const { cardId } = await params;
+
+		const card = await prisma.recognitionCard.findUnique({
+			where: { id: cardId },
+			select: { id: true },
+		});
```
Comment (chrisgen19) [inline app/api/recognition/[cardId]/interactions/route.ts:28]: The route returns comments and per-user reaction state for any existing `cardId` as long as the caller is logged in. Because the rest of the recognition UI restricts cards to the sender, recipient, or admins, this endpoint leaks interaction history for cards outside the caller's scope; a logged-in user can fetch `/api/recognition/{id}/interactions` for any shared card URL and see everyone's comments.

app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
chrisgen19: Fixed in commit 22f7b69.

Changed `enabled` to `false` by default so the query never fires on mount. The fetch is triggered lazily:
- **Desktop:** `onMouseEnter` on the interaction bar container — data is ready by the time the user clicks.
- **Mobile / first click:** `setFetchEnabled(true)` is called inside `handleReaction` and `handleCommentToggle` before any action fires.

When data isn't loaded yet and a reaction is clicked, the server action runs (which atomically handles add/remove) and the query is invalidated after — no stale state. The 6 ghost emoji buttons render immediately from the `REACTION_EMOJIS` constant without needing a fetch, so there's no visible loading state on page load.

app/(dashboard)/dashboard/recognition/_components/comment-thread.tsx
chrisgen19: Fixed in commit 22f7b69.

Changed the `onEdit` callback signature from `(commentId: string, newBody: string)` to `(updated: CardComment)`.

In `CommentItem.handleSaveEdit`, instead of passing just the body string, we now pass the full comment object constructed from the server response:
```ts
onEdit({
  ...comment,
  body: result.data.body,
  updatedAt: result.data.updatedAt.toISOString(),
});
```

In `CommentThread.handleEdit`, the map now replaces the entire comment object:
```ts
comments.map((c) => (c.id === updated.id ? updated : c))
```

The `· edited` badge and relative timestamp now update immediately after saving without waiting for a refetch.

lib/actions/interaction-actions.ts
chrisgen19: Not fixed — this is intentional by design.

The recognition feed on `/dashboard` already exposes **all** cards to every authenticated staff member via `GET /api/recognition?filter=all` (the default public feed). Cards are meant to be publicly visible to everyone in the organisation — that's the core purpose of the app.

Since any logged-in user can already see every card in the public feed, allowing them to react or comment on those same cards is consistent. Restricting reactions/comments to only the sender and recipient would make the interaction bar feel broken for the majority of cards shown in the feed.

If the scope ever changes (e.g. private cards, department-only visibility), this check should be revisited at that point.

app/api/recognition/[cardId]/interactions/route.ts
Comment (chrisgen19) [reply app/api/recognition/[cardId]/interactions/route.ts:28]: Not fixed — same reasoning as the comment on `interaction-actions.ts`.

`GET /api/recognition?filter=all` already returns all recognition cards to any authenticated user, and the dashboard feed renders them publicly. The interactions endpoint (`/api/recognition/{id}/interactions`) is scoped to the same visibility model — if a user can see a card in the feed, they can see its reactions and comments.

Locking this endpoint to sender/recipient only would mean the interactions bar silently fails for the majority of cards visible in the public feed. If per-card visibility rules are introduced in a future phase, the interactions endpoint should be updated to match them at that point.

lib/actions/interaction-actions.ts
chrisgen19: The current code still doesn't line up with that explanation. `GET /api/recognition` only falls back to `recipientId = session.user.id` for non-admins when no filter is supplied (`app/api/recognition/route.ts:43-45`), and `RecognitionDetailPage` still 404s unless the caller is the sender, recipient, or an admin (`app/(dashboard)/dashboard/recognition/[id]/page.tsx:94-98`). So a logged-in user who learns another card's UUID can still hit these mutations for a card the rest of the dashboard treats as out of scope. The sender/recipient/admin check still needs to run before the mutation proceeds.

app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	function handleCommentsChange(comments: CardComment[]) {
+		queryClient.setQueryData<{ success: boolean; data: CardInteractions }>(
+			["card-interactions", cardId],
+			(old) => {
+				if (!old?.data) return old;
+				return {
+					...old,
+					data: { ...old.data, comments, totalComments: comments.length },
```
chrisgen19: On slow connections, users can open comments and submit before the lazy `card-interactions` query has populated the cache. When that happens, `handleCommentsChange` returns `old` if `old?.data` is undefined, so the successful insert is discarded locally; if the in-flight GET misses the new row (or fails), the comment will not appear until a later refetch. Reactions already invalidate in this situation, but comments do not.

app/api/recognition/[cardId]/interactions/route.ts
Comment (chrisgen19) [reply app/api/recognition/[cardId]/interactions/route.ts:28]: The current visibility rules still don't match that explanation. In `app/api/recognition/route.ts`, non-admin callers without an explicit filter are scoped to `recipientId = session.user.id` (`43-45`), and the dashboard detail page still rejects any caller who is not the sender, recipient, or an admin (`app/(dashboard)/dashboard/recognition/[id]/page.tsx:94-98`). This endpoint is therefore exposing comments/reaction state for card IDs that the rest of the dashboard would treat as unauthorized, so it still needs the same card-level ACL check before returning the payload.

lib/actions/interaction-actions.ts
chrisgen19: Fixed in commit a5d1c28. You're right — I checked `GET /api/recognition?filter=all` instead of the no-filter path, which was wrong.

Added `senderId` and `recipientId` to the card select in both `toggleReactionAction` and `addCommentAction`, then enforce the same guard the detail page uses:

```ts
const isAdmin = hasMinRole(session.user.role as Role, 'ADMIN');
const isParticipant =
  card.senderId === session.user.id || card.recipientId === session.user.id;
if (!isParticipant && !isAdmin) {
  return { success: false, error: 'Forbidden' };
}
```

`editCommentAction` and `deleteCommentAction` are unaffected — they already enforce ownership/admin at the comment level, which implicitly scopes them to participants.

app/api/recognition/[cardId]/interactions/route.ts
Comment (chrisgen19) [reply app/api/recognition/[cardId]/interactions/route.ts:28]: Fixed in commit a5d1c28 — same fix as the server actions.

Extended the card select to include `senderId` and `recipientId`, then added the ACL check before returning any data:

```ts
const isAdmin = hasMinRole(session.user.role as Role, 'ADMIN');
const isParticipant =
  card.senderId === session.user.id || card.recipientId === session.user.id;
if (!isParticipant && !isAdmin) {
  return NextResponse.json({ success: false, error: 'Forbidden' }, { status: 403 });
}
```

The endpoint now returns 403 for any caller who is not the card's sender, recipient, or an admin — consistent with the detail page guard.

app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
chrisgen19: Fixed in commit a5d1c28.

When `old?.data` is undefined (lazy query hasn't fired yet), `handleCommentsChange` now seeds the cache from scratch instead of returning `undefined`:

```ts
if (!old?.data) {
  return {
    success: true,
    data: {
      reactions: REACTION_EMOJIS.map((emoji) => ({ emoji, count: 0, hasReacted: false })),
      comments,
      totalComments: comments.length,
    },
  };
}
```

This means a comment posted before the interactions query has populated always appears immediately in local state. When the lazy query does eventually fire (on hover or next interaction), it overwrites the seeded cache with the authoritative server data.

app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
```diff
@@ -176,17 +179,26 @@ export function RecognitionFeed({
...
+										{actions}
+									</div>
+								)}
+							</div>
+							<div className="px-5 pb-4">
+								<CardInteractionBar
+									cardId={card.id}
+									currentUserId={currentUserId}
+									isAdmin={isAdmin}
+								/>
```
chrisgen19: This now renders `CardInteractionBar` for every card, but `/api/recognition?filter=department` already returns cards that a staff user neither sent nor received. The new interactions route/actions only allow sender/recipient/admin, so on the dashboard's "My Department" tab those cards show a comment/reaction UI that can only fail with `403` (counts never load, and reacting/commenting toasts `Forbidden`). Gate the bar on participant/admin access or relax the backend ACL so this view does not advertise a broken action.

app/api/recognition/[cardId]/interactions/route.ts
```diff
@@ -0,0 +1,102 @@
...
+	{ params }: { params: Promise<{ cardId: string }> },
+) {
+	try {
+		const session = await getServerSession();
+		if (!session) {
+			return NextResponse.json(
+				{ success: false, error: "Unauthorized" },
+				{ status: 401 },
+			);
+		}
```
Comment (chrisgen19) [inline app/api/recognition/[cardId]/interactions/route.ts]: This route authenticates with `getServerSession()` instead of `requireSession()`, so a deactivated user who still has a valid session cookie can keep reading reactions/comments for card IDs they already know. The other recognition APIs go through `requireSession()` specifically to block inactive accounts, so this new endpoint weakens that boundary.

app/(dashboard)/dashboard/recognition/_components/comment-thread.tsx
```diff
@@ -0,0 +1,286 @@
...
+			const result = await addCommentAction(cardId, trimmed);
+			if (result.success) {
+				onCommentsChange([
+					...comments,
+					{
+						...result.data,
+						createdAt: result.data.createdAt.toISOString(),
+						updatedAt: result.data.updatedAt.toISOString(),
+					},
+				]);
```
chrisgen19, line 241: In the slow-connection case this patch is trying to fix, opening the thread starts the lazy `/api/recognition/[cardId]/interactions` request and this success path only appends to local cache. If that in-flight fetch resolves with the pre-insert snapshot, React Query will overwrite the seeded comment list and the just-posted comment disappears again until a later refetch. The mutation needs to cancel or invalidate `["card-interactions", cardId]`, not only call `onCommentsChange(...)`.

app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
chrisgen19: Fixed in commit c09f854.

Added per-card `isParticipant` / `canInteract` guard in `recognition-feed.tsx`:

```ts
const isParticipant =
  currentUserId === card.sender.id ||
  currentUserId === card.recipient.id;
const canInteract = isParticipant || isAdmin;
```

`CardInteractionBar` is now wrapped in `{canInteract && ...}` for both the physical-card and digital-card render paths, so the interaction bar is never rendered (and no 403 GET is ever fired) for users who are not the sender, recipient, or an admin.

app/api/recognition/[cardId]/interactions/route.ts
Comment (chrisgen19) [reply app/api/recognition/[cardId]/interactions/route.ts]: Fixed in commit c09f854.

Replaced the `getServerSession()` null-check pattern with `requireSession()`:

```ts
let session: Awaited<ReturnType<typeof requireSession>>;
try {
  session = await requireSession();
} catch {
  return NextResponse.json(
    { success: false, error: "Unauthorized" },
    { status: 401 },
  );
}
```

`requireSession()` throws on both missing and inactive/deactivated sessions, so deactivated users can no longer access interaction data from this endpoint.

app/(dashboard)/dashboard/recognition/_components/comment-thread.tsx
chrisgen19: Fixed in commit c09f854.

Made `handleCommentsChange` in `card-interaction-bar.tsx` async and added `cancelQueries` before `setQueryData`:

```ts
async function handleCommentsChange(comments: CardComment[]) {
  await queryClient.cancelQueries({ queryKey: ["card-interactions", cardId] });
  queryClient.setQueryData<{ success: boolean; data: CardInteractions }>(
    ["card-interactions", cardId],
    (old) => { ... },
  );
}
```

Any in-flight lazy GET for this card is cancelled before the seeded comment list is written to the cache, so a slow response can no longer overwrite the optimistic comment.

app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
```diff
@@ -0,0 +1,205 @@
...
+						hasReacted: adding,
+					};
+				});
+				return { ...old, data: { ...old.data, reactions } };
+			},
+		);
+
+		toggleReactionAction(cardId, emoji).then((result) => {
+			if (!result.success) {
+				queryClient.invalidateQueries({ queryKey: ["card-interactions", cardId] });
```
chrisgen19: The optimistic updater derives the new count from whatever reaction data was last cached, but on success we never reconcile with the server. When another participant has reacted since the last fetch, toggling from stale data leaves the count and `hasReacted` state incorrect until the 30s stale window expires or the page is reloaded, because only failures invalidate the query.

app/(dashboard)/dashboard/recognition/_components/comment-thread.tsx
```diff
@@ -0,0 +1,286 @@
...
+						</p>
+					)}
+				</div>
+				<div className="flex items-center gap-2 mt-0.5 px-1">
+					<span className="text-[10px] text-muted-foreground">
+						{timeAgo(comment.createdAt)}
+						{comment.updatedAt !== comment.createdAt && " · edited"}
+					</span>
+					{!isEditing && (canEdit || canDelete) && (
+						<div className="flex gap-1 opacity-0 group-hover/comment:opacity-100 transition-opacity">
```
chrisgen19: These controls are hidden with `opacity-0 group-hover/comment:opacity-100`, so they only become visible on hover. On touch devices, and for keyboard users who navigate without hover, owners/admins never get a visible affordance to edit or delete a comment, which makes those actions effectively unavailable in the new thread UI.

app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
```diff
@@ -0,0 +1,205 @@
...
+	currentUserId,
+	isAdmin,
+}: CardInteractionBarProps) {
+	const queryClient = useQueryClient();
+	const [showComments, setShowComments] = useState(false);
+	// Lazy: only fetch when user first hovers (desktop) or interacts (mobile)
+	const [fetchEnabled, setFetchEnabled] = useState(false);
+
+	const { data } = useQuery<{ success: boolean; data: CardInteractions }>({
+		queryKey: ["card-interactions", cardId],
```
chrisgen19: `/api/recognition/[cardId]/interactions` returns viewer-specific data (`hasReacted`) and its authorization also depends on who is signed in, but this cache key only uses `cardId`. Because the root `QueryClient` survives route changes, signing out and back in as another user on the same browser within `staleTime` can reuse the previous viewer's interaction state, so the next user can see the wrong reaction highlighted and act on stale counts.

app/(dashboard)/dashboard/recognition/_components/comment-thread.tsx
chrisgen19: `cancelQueries()` is not sufficient here because the query function never consumes React Query's abort signal. The original `/api/recognition/[cardId]/interactions` request will still resolve and write its stale pre-mutation payload into the cache, replacing the optimistic `comments` array. To actually prevent that overwrite, the fetch needs to accept the provided `signal` (or the mutation needs to force a fresh invalidate/refetch after success).

app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
chrisgen19: Fixed in commit 6cf144a.

The optimistic path now always invalidates after the server action resolves — both on success and failure:

```ts
toggleReactionAction(cardId, emoji).then((result) => {
  if (!result.success) {
    toast.error(result.error);
  }
  // Always invalidate to pick up concurrent reactions from other users
  queryClient.invalidateQueries({ queryKey: ["card-interactions", cardId, currentUserId] });
});
```

The optimistic update still fires immediately for snappy feedback. The background refetch that follows corrects any counts that drifted due to concurrent reactions from other users.

The same pattern was applied to the no-interactions (cold cache) path.

app/(dashboard)/dashboard/recognition/_components/comment-thread.tsx
chrisgen19: Fixed in commit 6cf144a.

Replaced `opacity-0 group-hover/comment:opacity-100` with the `@media(hover:hover)` variant:

```tsx
<div className="flex gap-1 [@media(hover:hover)]:opacity-0 [@media(hover:hover)]:group-hover/comment:opacity-100 transition-opacity">
```

On hover-capable devices (desktop), the controls still hide until hover. On touch/pointer-coarse devices where hover is not available, the media query doesn't match so the controls are always visible — same pattern used elsewhere in the codebase for `CardActions`.

app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
chrisgen19: Fixed in commit 6cf144a.

Added `currentUserId` as the third segment of every query key in `CardInteractionBar`:

```ts
queryKey: ["card-interactions", cardId, currentUserId],
```

Updated consistently across `useQuery`, all `setQueryData`, `invalidateQueries`, and `cancelQueries` calls. Each user now gets an isolated cache entry, so signing in as a different user within the staleTime window always starts with a clean slate.

app/(dashboard)/dashboard/recognition/_components/comment-thread.tsx
chrisgen19: Fixed in commit 6cf144a.

The `queryFn` now accepts and forwards React Query's abort signal to `fetch`:

```ts
queryFn: async ({ signal }) => {
  const res = await fetch(`/api/recognition/${cardId}/interactions`, { signal });
  if (!res.ok) throw new Error("Failed to fetch interactions");
  return res.json();
},
```

When `cancelQueries` is called in `handleCommentsChange`, the abort signal fires and the browser actually aborts the in-flight request. The response can no longer resolve and overwrite the optimistically seeded comment list.

chrisgen19:
## Fix summary — all rounds

### Round 1 (commit 22f7b69)
- **N+1 fetch on mount** (`card-interaction-bar.tsx`) — set `enabled: false` by default; fetch triggers lazily on `onMouseEnter` (desktop) or first interaction (mobile). Ghost buttons render from `REACTION_EMOJIS` constant so there is no visible loading state.
- **Stale `updatedAt` after edit** (`comment-thread.tsx`) — changed `onEdit` callback to pass the full `CardComment` object from the server response (including `updatedAt`). The `· edited` badge and timestamp now update immediately.

### Round 2 (commit a5d1c28)
- **Missing ACL on server actions** (`interaction-actions.ts`) — added `senderId`/`recipientId` to the card select in `toggleReactionAction` and `addCommentAction`; enforce sender/recipient/admin guard before mutating.
- **Missing ACL on GET route** (`interactions/route.ts`) — same guard added; endpoint now returns 403 for callers outside the card's sender, recipient, or admin role.
- **Cold-cache comment lost** (`card-interaction-bar.tsx`) — `handleCommentsChange` now seeds the full cache structure when `old?.data` is undefined, so a comment posted before the lazy query fires always appears immediately.

### Round 3 (commit c09f854)
- **Department feed 403 UI** (`recognition-feed.tsx`) — added per-card `isParticipant`/`canInteract` check; `CardInteractionBar` is gated with `{canInteract && …}` in both render paths so it is never rendered for non-participants.
- **Deactivated user bypass** (`interactions/route.ts`) — replaced `getServerSession()` null-check with `requireSession()` two-try-catch pattern; inactive sessions are now rejected with 401.
- **In-flight fetch overwrites seeded comment** (`card-interaction-bar.tsx`) — made `handleCommentsChange` async and added `await cancelQueries` before `setQueryData`.

### Round 4 (commit 6cf144a)
- **Stale reaction counts on success** (`card-interaction-bar.tsx`) — `invalidateQueries` now fires on both success and failure after `toggleReactionAction`, so concurrent reactions from other users are reconciled after every toggle.
- **Touch/keyboard edit/delete inaccessible** (`comment-thread.tsx`) — replaced `opacity-0 group-hover` with `[@media(hover:hover)]:opacity-0 [@media(hover:hover)]:group-hover` variants; controls are always visible on touch devices and for keyboard users.
- **User-scoped cache key** (`card-interaction-bar.tsx`) — added `currentUserId` as the third key segment everywhere (`["card-interactions", cardId, currentUserId]`); each user gets an isolated cache entry so `hasReacted` state is never leaked across users within the stale window.
- **`cancelQueries` didn't abort fetch** (`card-interaction-bar.tsx`) — the `queryFn` now accepts `{ signal }` and forwards it to `fetch()`; the browser actually cancels the in-flight request when `cancelQueries` fires.

app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
```diff
@@ -0,0 +1,207 @@
...
+					return {
+						success: true,
+						data: {
+							reactions: REACTION_EMOJIS.map((emoji) => ({
+								emoji,
+								count: 0,
+								hasReacted: false,
+							})),
+							comments,
+							totalComments: comments.length,
```
chrisgen19, line 122: If the first `GET /api/recognition/[cardId]/interactions` request is still in flight and the user posts a comment immediately, this `!old?.data` fallback seeds the cache from the local `comments` array plus zero-count reactions. In that slow-network path, any pre-existing comments or reactions disappear from the thread, and because the in-flight query was canceled without a follow-up refetch, the incorrect state can stick until the query goes stale or the page reloads.

lib/actions/interaction-actions.ts
```diff
@@ -0,0 +1,193 @@
...
+
+		const existing = await prisma.cardReaction.findUnique({
+			where: {
+				cardId_userId_emoji: {
+					cardId,
+					userId: session.user.id,
+					emoji,
+				},
+			},
+		});
```
chrisgen19, line 50: This toggle is implemented as a `findUnique` followed by `create` or `delete`, so two quick taps on the same emoji can race against each other. Both requests can observe the same pre-toggle state, which either leaves the server state opposite to the last user tap or throws a Prisma unique or P2025 error back to the UI. Because the client does not disable the button while a toggle is pending, this race is user-reachable.

app/api/recognition/route.ts
```diff
@@ -63,8 +63,15 @@ export async function GET(request: NextRequest) {
...
 				},
 			},
+			_count: {
+				select: { reactions: true, comments: true },
+			},
 		};
 
+		const mapCounts = <T extends { _count: { reactions: number; comments: number } }>(
+			{ _count, ...card }: T,
+		) => ({ ...card, interactionCounts: _count });
```
chrisgen19: For department-feed requests, users who are neither the sender nor recipient now receive `interactionCounts` for cards they cannot open the interaction API for. That leaks participant-only activity through `/api/recognition`, even though `/api/recognition/[cardId]/interactions` returns `403` for the same user.

lib/actions/interaction-actions.ts
```diff
@@ -0,0 +1,193 @@
...
+		const session = await requireSession();
+		const parsedBody = commentBodySchema.safeParse(body);
+		if (!parsedBody.success) {
+			return { success: false as const, error: parsedBody.error.issues[0].message };
+		}
+
+		const comment = await prisma.cardComment.findUnique({
+			where: { id: commentId },
+			select: { userId: true },
+		});
```
chrisgen19, line 145: `editCommentAction` and `deleteCommentAction` only verify comment ownership or admin status here, not whether the caller is still allowed to access the card. If a sender changes the recipient after comments already exist, a former recipient can still edit or delete an old comment from an already-open tab even though `GET /interactions` and `addCommentAction` would now return `403`.

app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
chrisgen19: Fixed in commit ad79d7e.

When `!old?.data`, the seed now sets a `seededFromScratch` flag. After `setQueryData`, if the flag is true, `invalidateQueries` fires so a background refetch reconciles any pre-existing reactions/comments the seed didn't know about. The seeded data stays visible immediately while the refetch runs in the background.

lib/actions/interaction-actions.ts
chrisgen19: Fixed in commit ad79d7e.

**Client:** Added a `pendingToggles` ref (`Set<string>`) — `handleReaction` returns early if that emoji is already in flight, so double-taps never reach the server.

**Server:** Replaced the non-atomic `findUnique` → `create`/`delete` with:
1. `deleteMany` — if count > 0, reaction removed
2. `create` — if no existing reaction, create one
3. Catch P2002 (unique constraint) — concurrent toggle already created it, so delete it instead

This ensures the server converges to a consistent state even if two requests slip through.

app/api/recognition/route.ts
chrisgen19: Fixed in commit ad79d7e.

Updated `mapCounts` to check `senderId`/`recipientId`/`isAdmin` — non-participant cards now return `interactionCounts: null`, matching the 403 behavior of `/api/recognition/[cardId]/interactions`.

lib/actions/interaction-actions.ts
chrisgen19: Fixed in commit ad79d7e.

Both `editCommentAction` and `deleteCommentAction` now include `card: { select: { senderId, recipientId } }` in the comment query and verify the caller is a participant (sender/recipient) or admin before allowing the mutation. Returns `"Forbidden"` if the caller no longer has card access.

chrisgen19:
## Fix summary — Round 5 (commits ad79d7e, 11eef44)

### Round 5a (commit ad79d7e) — remaining review issues
- **Stale cache seed** (`card-interaction-bar.tsx`) — after seeding cache from scratch (`!old?.data`), `invalidateQueries` now fires so a background refetch reconciles pre-existing reactions/comments.
- **Toggle race condition** (`card-interaction-bar.tsx` + `interaction-actions.ts`) — client-side `pendingToggles` ref prevents double-taps reaching the server; server-side replaced non-atomic `findUnique`→`create`/`delete` with `deleteMany`→`create` + P2002 catch.
- **Department feed leaks counts** (`recognition/route.ts`) — `mapCounts` returns `interactionCounts: null` for non-participant cards.
- **Edit/delete missing card ACL** (`interaction-actions.ts`) — both actions now include `card: { select: { senderId, recipientId } }` and verify caller is participant/admin before mutating.

### Round 5b (commit 11eef44) — emoji counts hidden until hover
- **Root cause:** feed API had total `interactionCounts` but no per-emoji breakdown; `CardInteractionBar` showed ghost buttons with zero counts until the lazy fetch fired on hover.
- **Fix:** feed API now includes `reactionSummary` (per-emoji counts + `hasReacted`) via two batch queries (`groupBy` + user reactions lookup). `CardInteractionBar` accepts `initialReactions` prop and renders actual counts immediately. Ghost buttons only show for emojis with zero reactions. Lazy fetch still fires on hover/interaction for full detail reconciliation.

chrisgen19:
## Fix summary — Round 6 (commit 9f94b5b)

Addresses issues from code review:

1. **Seed uses `initialReactions`** — `handleCommentsChange` now populates the cache seed with real reaction counts from `initialReactions` instead of all-zero, eliminating the flash when a comment is submitted before hover.

2. **Delete confirmation** — Comment delete now uses a two-click pattern: first click shows "Delete?" text + cancel (X) button; second click performs the actual delete. Prevents accidental deletions on small 11px touch targets.

3. **"Me" → real initials** — Comment input avatar now shows the current user's actual initials via `useSession()` from `lib/auth-client.ts`, consistent with all other avatars in the app.

4. **Split `useTransition`** — `CommentItem` now uses separate `isEditPending`/`isDeletePending` transitions so editing a comment doesn't disable the delete button and vice versa.

**Skipped (verified as non-issues):**
- Issue #3 (`_count` type) — Prisma 7's `groupBy` with `_count: true` resolves to `number` via generic conditional type. Confirmed in generated types.
- Issue #2 (paginated admin path) — Admin table doesn't render `CardInteractionBar`, so missing `reactionSummary` is a non-issue.

PR #22, opened 2026-04-15
feat(recognition): configurable Most Recognized limit in admin settings

### LINKED ISSUE #21: feat(recognition): configurable 'Most Recognized' limit in admin settings
## Summary

The "Most Recognized" leaderboard in the Recognition Stats widget on the dashboard currently hardcodes a top 3 limit. This should be:

1. **Increased to top 10 by default** — showing only 3 doesn't give enough visibility
2. **Configurable by admins** — admins should be able to set the limit (1–50) from `/dashboard/admin-settings`

## Implementation

- Add a `top_recognized_limit` key to `AppSetting` (default: `10`)
- Add server actions to get/update the setting (ADMIN role required)
- Update `GET /api/recognition/stats` to read the setting instead of hardcoding `take: 3`
- Build a Recognition Settings panel on the admin settings page with a select dropdown
- Update the stats widget to render the dynamic list

## Acceptance Criteria

- [ ] Dashboard "Most Recognized" shows top 10 by default
- [ ] Admins can change the limit from admin settings (range: 1–50)
- [ ] Non-admins cannot access or change the setting
- [ ] Stats widget adapts to whatever limit is configured

chrisgen19:
## Summary

- Increases the "Most Recognized" leaderboard from hardcoded top 3 to a **default of 10**
- Adds an admin-configurable limit (1–50) via a new **Recognition Settings** panel on `/dashboard/admin-settings`
- Uses the existing `AppSetting` key-value table (`top_recognized_limit` key) with a 30s in-memory cache
- Stats widget now shows **rank numbers** and a **scrollable list** to handle larger sets

## Changes

| File | What changed |
|---|---|
| `lib/actions/settings-actions.ts` | Added `getTopRecognizedLimit()` and `updateTopRecognizedLimit()` server actions |
| `app/api/recognition/stats/route.ts` | Replaced `take: 3` with dynamic `topLimit` from AppSetting |
| `admin-settings/page.tsx` | Replaced placeholder with server component that fetches current limit |
| `admin-settings/_components/recognition-settings.tsx` | New client component with select dropdown (ADMIN role required) |
| `dashboard/_components/stats-widget.tsx` | Added rank numbers, `truncate` on names, scrollable `max-h-80` container |
| `prisma/seed.ts` | Seeds `top_recognized_limit = 10` for dev environments |

## Test plan

- [ ] Log in as ADMIN, go to `/dashboard/admin-settings` — verify Recognition Settings panel renders with current limit
- [ ] Change the limit via the dropdown — verify toast confirmation and value persists on page reload
- [ ] Go to `/dashboard` — verify "Most Recognized" shows the configured number of recipients
- [ ] Log in as STAFF — verify `/dashboard/admin-settings` redirects to `/dashboard`
- [ ] With no `top_recognized_limit` row in DB, verify default of 10 is used

Closes #21

### DIFF
--- a/app/(dashboard)/dashboard/admin-settings/_components/recognition-settings.tsx
+++ b/app/(dashboard)/dashboard/admin-settings/_components/recognition-settings.tsx
@@ -1,0 +1,84 @@
+"use client";
+
+import { useState, useTransition } from "react";
+import { toast } from "sonner";
+import {
+	Select,
+	SelectContent,
+	SelectItem,
+	SelectTrigger,
+	SelectValue,
+} from "@/components/ui/select";
+import { updateTopRecognizedLimit } from "@/lib/actions/settings-actions";
+
+const LIMIT_OPTIONS = [1, 3, 5, 10, 15, 20, 25, 30, 40, 50];
+
+export function RecognitionSettingsPanel({
+	initialLimit,
+}: {
+	initialLimit: number;
+}) {
+	const [limit, setLimit] = useState(initialLimit);
+	const [isPending, startTransition] = useTransition();
+
+	function handleChange(newValue: number) {
+		const previous = limit;
+		setLimit(newValue);
+
+		startTransition(async () => {
+			const result = await updateTopRecognizedLimit(newValue);
+			if (!result.success) {
+				setLimit(previous);
+				toast.error(result.error ?? "Failed to update setting");
+			} else {
+				toast.success(`Most Recognized limit updated to ${newValue}`);
+			}
+		});
+	}
+
+	return (
+		<div className="rounded-[2rem] border border-gray-200/60 dark:border-white/10 bg-card shadow-[0_2px_20px_-4px_rgba(0,0,0,0.05)] overflow-hidden">
+			<div className="px-8 pt-8 pb-2">
+				<h3 className="text-[1.5rem] leading-tight font-medium text-foreground tracking-tight">
+					Recognition Settings
+				</h3>
+				<p className="mt-2 text-sm text-muted-foreground">
+					Configure how recognition stats are displayed on the dashboard.
+				</p>
+			</div>
+
+			<div className="px-8 py-6">
+				<div className="flex items-center justify-between rounded-2xl border border-gray-200 dark:border-white/10 p-5">
+					<div>
+						<p className="text-sm font-medium text-foreground">
+							Most Recognized Limit
+						</p>
+						<p className="text-xs text-muted-foreground">
+							Number of top recipients shown in the dashboard stats
+							widget.
+						</p>
+					</div>
+
+					<Select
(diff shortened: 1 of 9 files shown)


lib/actions/settings-actions.ts
```diff
@@ -13,6 +13,14 @@ let cachedSettings: OAuthSettings | null = null;
 let cacheExpiry = 0;
 const CACHE_TTL_MS = 30_000;
 
+const TOP_RECOGNIZED_KEY = "top_recognized_limit";
+const TOP_RECOGNIZED_DEFAULT = 10;
+const TOP_RECOGNIZED_MIN = 1;
+const TOP_RECOGNIZED_MAX = 50;
+
+let cachedTopLimit: number | null = null;
+let topLimitCacheExpiry = 0;
```
chrisgen19, line 22: After `/api/recognition/stats` has been hit once, this in-memory cache can keep the old limit alive even after an admin saves a new one. `getTopRecognizedLimit()` is imported by both the server action and the stats route, and invalidating a module-local variable here does not reliably invalidate the copy used by the other server entrypoint, so the dashboard can continue returning the previous number of recipients for up to 30 seconds.

app/(dashboard)/dashboard/admin-settings/_components/recognition-settings.tsx
```diff
@@ -0,0 +1,84 @@
...
+}) {
+	const [limit, setLimit] = useState(initialLimit);
+	const [isPending, startTransition] = useTransition();
+
+	function handleChange(newValue: number) {
+		const previous = limit;
+		setLimit(newValue);
+
+		startTransition(async () => {
+			const result = await updateTopRecognizedLimit(newValue);
```
chrisgen19: If `updateTopRecognizedLimit()` throws instead of returning `{ success: false }` (for example during a transient Prisma or DB failure), this awaited call rejects before the rollback branch runs. In that case the select stays on the optimistic value and no error toast is shown even though the setting was never persisted.

app/(dashboard)/dashboard/admin-settings/_components/recognition-settings.tsx
```diff
@@ -0,0 +1,84 @@
...
+			if (!result.success) {
+				setLimit(previous);
+				toast.error(result.error ?? "Failed to update setting");
+			} else {
+				toast.success(`Most Recognized limit updated to ${newValue}`);
+			}
```
chrisgen19, line 43: If an admin has already opened `/dashboard` in the same SPA session, the root `QueryClient` can still consider `["recognition-stats"]` fresh for 30 seconds. After this save succeeds, navigating back to the dashboard will reuse the cached stats and still show the old recipient count until that TTL expires, which makes the setting look ineffective even though the write succeeded.

PR #25, opened 2026-04-16
fix(recognition): allow FROM/DATE box to grow so name text isn't clipped

chrisgen19:
## Summary
- Regression introduced by #24: names with descenders (e.g. "Tamer Abdelatty") render with their bottoms clipped inside the FROM box.
- Root cause: `s.fromDateH` is a fixed height (`h-12 md:h-14` at normal) that's a few px shorter than `padding + label + margin + value line-height`. Before #24 the text silently overflowed the box; the `overflow-hidden` I added to `FitText` as an ellipsis safety net now makes that overflow visible as clipping.
- Fix: change `h-*` → `min-h-*` in all three size configs' `fromDateH` so boxes can grow to fit content while keeping the original minimum. Flex-row `items-stretch` keeps FROM and DATE matched in height.

## Test plan
- [ ] "Tamer Abdelatty" in FROM renders fully (no clipped descender).
- [ ] "Christian Diomampo" and other long names stay on a single line (FitText still in play).
- [ ] FROM and DATE boxes visually match in height at compact, normal, expanded card sizes.
- [ ] Short names and dates keep the minimum height — no visible shrinkage vs. before.


chrisgen19:
The earlier review output no longer matches the current `#25` diff: this PR already switches the fixed `h-*` sender/date boxes to `min-h-*`, and it already applies `FitText` on the detail/share card renderers. The remaining issue I still see is inherited from `main` via `components/shared/fit-text.tsx`: `measure()` clears the inline `font-size` before recomputing it, but if the next computed size is unchanged React skips `setFontSize(...)`. On a later resize that still needs the same shrunken size, the inline style can be dropped permanently and the text jumps back to the base CSS size, so long names can overflow again after window resizes or similar layout changes. I'm leaving this as a top-level note here because `fit-text.tsx` is not part of the `#25` diff, so it can't be attached as an inline review comment on this PR.

chrisgen19:
the detail/share card pages, and the new measurement logic can drop the applied shrink size on later resizes. Those are reachable user-facing issues, so the patch is not fully correct.

Full review comments:

- [P2] Reapply the shrunken size when a remeasure returns the same value — /home/chrisgen19/projects/access-group-staff/components/shared/fit-text.tsx:29-30
  `measure()` clears the inline `font-size` before recomputing it, but React will skip `setFontSize(...)` when the next value is unchanged. That means any later resize that still leaves a long name pinned at the same `minFontSize` removes the inline style permanently, so the text jumps back to the base CSS size and can overflow again after a window resize or similar layout change.
- [P2] Let the FROM field grow before wrapping it in `FitText` — /home/chrisgen19/projects/access-group-staff/app/(dashboard)/dashboard/recognition/_components/recognition-card-mini.tsx:232-234
  The sender box is still constrained by the fixed `h-*` height in `s.fromDateH`, but `FitText` now adds `overflow-hidden` to the value line. For names with descenders such as `Tamer Abdelatty`, the line is a few pixels taller than the available vertical space, so the bottom of `y/g/p` gets visibly clipped in the physical card. Using a minimum height here avoids introducing that regression.
- [P2] Apply the long-name fix to the other card renderers too — /home/chrisgen19/projects/access-group-staff/app/(dashboard)/dashboard/recognition/_components/recognition-card-mini.tsx:162-164
  This patch only wires `FitText` into `RecognitionCardMini`. The same fixed-height name fields in /home/chrisgen19/projects/access-group-staff/app/(dashboard)/dashboard/recognition/[id]/page.tsx and /home/chrisgen19/projects/access-group-staff/app/recognition/[id]/page.tsx still render plain spans, so long sender/recipient names continue to overflow on the detail view and on public/shared cards even after this PR lands.

Issue #33, opened 2026-04-17
Leaderboard: monthly reset + configurable reveal window + archive


chrisgen19:
## Problem

The Recognition Stats widget on `/dashboard` currently renders an **always-visible, all-time** "Most Recognized" leaderboard. This causes two problems:

1. **Gaming** — users can spam recognition cards to climb the rankings at any time.
2. **No fresh start** — the leaderboard never resets, so early-active users dominate indefinitely and new users have no realistic path to the podium.

## Goal

Turn Most Recognized into a **monthly contest with a configurable reveal window**, and **archive** each month's results for historical record. Personal stats (Cards Sent / Cards Received / This Month) remain always visible — only the ranked leaderboard is gated.

## Decisions (already agreed)

- **Scope** — leaderboard becomes monthly (resets each calendar month, Asia/Manila). Personal counts stay all-time.
- **Timezone** — all month boundaries and reveal windows use `Asia/Manila`.
- **Admin bypass** — none. Admins see the same locked state as users (strict fairness).
- **Hidden state UI** — locked card with lock icon + reveal date range. Example: _"Rankings revealed Apr 24 – Apr 30"_.
- **Visibility modes**:
  - `always` — current behavior, no gating (default).
  - `last_n_days_of_month` — show only during last N days (1–14) of the current calendar month in Asia/Manila.
  - `custom_range` — one-off start/end dates picked by admin.
- **Snapshot trigger** — **lazy**: first request after month rollover checks if prior month is archived; if not, writes the snapshot. No cron needed.
- **History UI** — `/dashboard/leaderboard` (new page, everyone). Month picker shows archived months.

### PR 1 — Backend + visibility gating (this PR)

- [ ] Add `MonthlyLeaderboardSnapshot` Prisma model (month unique key, recipients JSON, topLimit, snapshotAt)
- [ ] Add `lib/leaderboard/month.ts` — Asia/Manila month boundary helpers
- [ ] Add `lib/leaderboard/visibility.ts` — compute `{ visible, revealStart, revealEnd }` across all 3 modes
- [ ] Add `lib/leaderboard/snapshot.ts` — `maybeSnapshotPreviousMonth()` lazy archiver
- [ ] Extend `lib/actions/settings-actions.ts` with visibility settings (mode, reveal days, custom start/end)
- [ ] Update `/api/recognition/stats` route to:
  - Run lazy snapshot on every request
  - Scope leaderboard to current Manila month
[...]

PR #34, opened 2026-04-17
feat(leaderboard): monthly reset + reveal window + archive (1/2)

chrisgen19:
## Summary

Part 1 of 2 for #33. Turns the Most Recognized leaderboard into a monthly contest with a configurable reveal window, and starts archiving each month's results. Personal stats (Cards Sent / Received / This Month) remain always visible — only the ranked leaderboard is gated.

- **Monthly reset** — leaderboard now scopes to the current Asia/Manila calendar month instead of all-time.
- **Configurable reveal window** — admin-settings lets admins pick `always`, `last N days of month` (N = 1–14), or a one-off `custom range`.
- **Locked state UI** — outside the window, the widget shows a lock icon and the reveal date range (e.g. _"Rankings revealed Apr 24 – Apr 30"_). Same UX for everyone — no admin bypass.
- **Lazy archive** — first `/api/recognition/stats` request after month rollover writes a `MonthlyLeaderboardSnapshot` for the prior month. No cron needed; upsert is idempotent so races are benign.

PR 2 will add the user-facing `/dashboard/leaderboard` history page that reads from these snapshots.

## Files touched

- `prisma/schema.prisma` — new `MonthlyLeaderboardSnapshot` model (keyed on `month` = `YYYY-MM`, JSON recipients payload)
- `lib/leaderboard/month.ts` — Asia/Manila boundary helpers + month-key parsing
- `lib/leaderboard/visibility.ts` — `computeLeaderboardVisibility()` across all 3 modes
- `lib/leaderboard/snapshot.ts` — `maybeSnapshotPreviousMonth()` lazy archiver
- `lib/actions/settings-actions.ts` — 4 new AppSetting keys + `get/updateLeaderboardVisibilitySettings`
- `app/api/recognition/stats/route.ts` — month-scoped leaderboard, visibility gating, snapshot call
- `app/(dashboard)/dashboard/_components/stats-widget.tsx` — locked card rendering
- `app/(dashboard)/dashboard/admin-settings/` — new visibility panel

## Deployment notes

- Schema is additive (one new table, no existing columns changed). Run `bunx prisma db push` (dev) or `bunx prisma migrate deploy` (prod) before merging.
- No new npm packages.
- Timezone is hardcoded to Asia/Manila as agreed.

## Test plan

- [ ] Run `bunx prisma db push` locally to apply the schema
- [ ] Visit `/dashboard/admin-settings` as ADMIN — verify the new "Leaderboard Visibility" panel renders, changing mode persists across reload
- [ ] Set mode to `always` — `/dashboard` shows Most Recognized (current-month only)
- [ ] Set mode to `last_n_days_of_month` with N = 1 on the last day of a month — widget shows the list; set N = 1 mid-month — widget shows the locked card with correct reveal range
- [ ] Set mode to `custom_range` with a future start date — widget shows locked card with that range
- [ ] Set mode to `custom_range` with dates in the past — widget shows "Rankings hidden" (no range)
- [ ] Verify non-admin users see the same locked state (no admin bypass)
- [ ] After a month rollover, inspect `monthly_leaderboard_snapshots` table — previous month should be written on first stats fetch
- [ ] `bun lint` + `bun run build` pass locally

Closes in part #33.

### DIFF
--- a/app/(dashboard)/dashboard/_components/stats-widget.tsx
+++ b/app/(dashboard)/dashboard/_components/stats-widget.tsx
@@ -1,9 +1,9 @@
 "use client";
 
 import { useQuery } from "@tanstack/react-query";
-import { Send, Inbox, Calendar, Trophy, Medal } from "lucide-react";
-import { cn } from "@/lib/utils";
+import { Calendar, Heart, Inbox, Lock, Medal, Send, Trophy } from "lucide-react";
 import { UserAvatar } from "@/components/shared/user-avatar";
+import { cn } from "@/lib/utils";
 
 const PODIUM_STYLES = [
 	{
@@ -29,6 +29,15 @@ const PODIUM_STYLES = [
 	},
 ] as const;
 
+type LeaderboardVisibilityMode = "always" | "last_n_days_of_month" | "custom_range";
+
+interface LeaderboardVisibility {
+	visible: boolean;
+	mode: LeaderboardVisibilityMode;
+	revealStart: string | null;
+	revealEnd: string | null;
+}
+
 interface StatsData {
 	sent: number;
 	received: number;
@@ -39,8 +48,21 @@ interface StatsData {
 		avatar: string | null;
 		count: number;
 	}[];
+	leaderboardVisibility: LeaderboardVisibility;
 }
 
+const REVEAL_DATE_FORMATTER = new Intl.DateTimeFormat("en-AU", {
+	timeZone: "Asia/Manila",
+	month: "short",
+	day: "numeric",
+});
+
+function formatRevealRange(startIso: string, endIso: string): string {
+	const start = new Date(startIso);
+	// revealEnd is exclusive — subtract 1ms to display the inclusive last day
+	const lastDay = new Date(new Date(endIso).getTime() - 1);
+	return `${REVEAL_DATE_FORMATTER.format(start)} – ${REVEAL_DATE_FORMATTER.format(lastDay)}`;
+}
 
 function StatItem({
 	icon: Icon,
@@ -66,7 +88,11 @@ function StatItem({
 
 function StatsWidgetSkeleton() {
 	return (
-		<div className="rounded-[2rem] border border-gray-200/60 dark:border-white/10 bg-card p-6 shadow-[0_2px_20px_-4px_rgba(0,0,0,0.03)] animate-pulse space-y-6 h-full" aria-busy="true" aria-label="Loading recognition stats">
+		<div
+			className="rounded-[2rem] border border-gray-200/60 dark:border-white/10 bg-card p-6 shadow-[0_2px_20px_-4px_rgba(0,0,0,0.03)] animate-pulse space-y-6 h-full"
+			aria-busy="true"
+			aria-label="Loading recognition stats"
+		>
 			<div className="h-5 w-32 bg-gray-200 dark:bg-white/10 rounded" />
 			<div className="grid grid-cols-3 gap-4">
 				{[1, 2, 3].map((i) => (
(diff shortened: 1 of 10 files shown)


app/(dashboard)/dashboard/_components/stats-widget.tsx
```diff
@@ -91,6 +163,46 @@ function StatsWidgetSkeleton() {
 	);
 }
 
+function LockedLeaderboard({ visibility }: { visibility: LeaderboardVisibility }) {
+	const hasRange = visibility.revealStart && visibility.revealEnd;
+	const rangeLabel = hasRange
+		? formatRevealRange(visibility.revealStart as string, visibility.revealEnd as string)
+		: null;
+	const msRemaining = useCountdown(visibility.revealStart);
```
chrisgen19: [P1] `LockedLeaderboard` only starts a timer for `revealStart`, so once the leaderboard is already visible there is no corresponding refresh at `revealEnd`. If someone leaves the dashboard open through the end of a custom range, or across Manila month-end in `last_n_days_of_month` mode, the cached `recognition-stats` response keeps rendering rankings after they should be hidden/reset until the tab refocuses or some other action invalidates the query.

app/(dashboard)/dashboard/_components/stats-widget.tsx
```diff
@@ -39,8 +49,66 @@ interface StatsData {
...
+		function tick() {
+			const remaining = target - Date.now();
+			setMsRemaining(remaining);
+			if (remaining <= 0) {
+				queryClient.invalidateQueries({ queryKey: ["recognition-stats"] });
+			}
+		}
+
+		tick();
+		const intervalMs = target - Date.now() > 60 * 60 * 1000 ? 60_000 : 1_000;
```
chrisgen19: [P2] When the widget mounts more than an hour before `revealStart`, `intervalMs` is fixed to 60 seconds relative to mount time and never switches to 1-second updates later. For midnight reveal times, the next tick can therefore land up to 59 seconds after the boundary, so long-lived tabs can stay locked after rankings should already be visible and the promised seconds-level countdown never appears.

lib/leaderboard/snapshot.ts
```diff
@@ -0,0 +1,71 @@
...
+export async function maybeSnapshotPreviousMonth(now: Date = new Date()): Promise<void> {
+	const prevKey = getPreviousMonthKey(now);
+
+	const existing = await prisma.monthlyLeaderboardSnapshot.findUnique({
+		where: { month: prevKey },
+		select: { id: true },
+	});
+	if (existing) return;
+
+	const topLimit = await getTopRecognizedLimit();
```
chrisgen19: [P2] Because snapshotting is lazy, this reads today's `top_recognized_limit`, not the limit that was active while `prevKey` was live. If an admin changes the limit on the 1st before anyone loads the dashboard, the archived snapshot for the previous month is truncated or expanded using the new cutoff, and the stored `topLimit` no longer matches what users actually saw that month.

chrisgen19 (commented):
[P1] The new reveal-window behavior is not fully enforced for long-lived dashboard sessions, and the lazy snapshot logic can archive a previous month with the wrong leaderboard cutoff. Those issues make the time-based leaderboard feature observably incorrect.

Inline comments posted:
- `stats-widget.tsx`: invalidate visible rankings when the reveal window ends
- `stats-widget.tsx`: re-arm the countdown when the remaining time drops below an hour
- `snapshot.ts`: snapshot the previous month before applying the new top limit

app/(dashboard)/dashboard/_components/stats-widget.tsx
chrisgen19: Addressed in dcf8a39. Lifted the countdown into `StatsWidget` so it always watches the next boundary — `revealEnd` when visible, `revealStart` when locked — invalidating the stats query whichever way the boundary flips. Long-lived tabs now hide the rankings at `revealEnd` instead of waiting for focus/refetch.

app/(dashboard)/dashboard/_components/stats-widget.tsx
chrisgen19: Addressed in dcf8a39. Switched `useCountdown` to recursive `setTimeout`, so `nextDelay` is recomputed on every tick. A tab mounted hours out will now upgrade from 60s to 1s as the boundary approaches, and the zero-crossing fires within 1s of the actual boundary instead of up to 59s late. Also added a small guard to skip the invalidation if the component mounts with `remaining <= 0` (server state is already authoritative in that case).

lib/leaderboard/snapshot.ts
chrisgen19: Addressed in dcf8a39. The snapshot now captures up to `SNAPSHOT_TOP_N = 50` (matching `TOP_RECOGNIZED_MAX`) rather than the current `top_recognized_limit`. Archived snapshots are now authoritative and independent of later admin changes; the history view (PR 2) will trim to the current limit at display time.

PR #39, opened 2026-04-17
feat(users): port recognition table + filter pattern to Staff Directory

### LINKED ISSUE #38: Apply recognition table/filter pattern to Staff Directory
## Summary
Port the table + filter experience used on `/dashboard/recognition/all` to `/dashboard/users` so the Staff Directory gets the same search, chip filters, server-side pagination, and CSV export.

## Motivation
The recognition all-cards page (`RecognitionTable` + `RecognitionFilterBar`) provides a much better browsing experience than the current ad-hoc users table: debounced search, filter chips, paginated fetches, and an export button. Staff Directory should match so admins can find/filter/export users consistently.

### UI
- Swap current users `<table>` for shadcn `Table` primitives styled like the recognition table (rounded card shell, muted header row, hover rows, per-row action buttons).
- Introduce a new `UserFilterBar` mirroring `RecognitionFilterBar`:
  - Search input (debounced 300ms)
  - **Role** chips: `STAFF` / `ADMIN` / `SUPERADMIN` (SUPERADMIN viewer only — matches existing role-badge gating)
  - **Status** chips: Active / Inactive
  - **Department** dropdown (populated via `getDepartmentsAction`)
  - **Branch** dropdown (`ISO` / `PERTH` enum values)
  - Clear filters + Export CSV
- Server-side pagination (page size 20) with a footer like recognition's (`showing X–Y of N`, prev/next, page indicator).
- Preserve existing row actions: View, Edit, Activate/Deactivate.

### API
- Extend `GET /api/users` to accept `paginated`, `page`, `pageSize`, `search`, `roles`, `statuses`, `departmentId`, `branch`, and `export=true`.
- Paginated response matches `/api/recognition` shape: `{ success, data, pagination: { page, pageSize, total, totalPages } }`.
- Export returns `{ success, data }` up to a sane cap (10k).

### CSV
- Add a `generateUserCsv` helper — columns: First Name, Last Name, Email, Role, Department, Branch, Position, Status.

## Non-goals
- No changes to user create/edit flows.
- No new permissions changes beyond preserving current role-based column visibility.

## Acceptance
- `/dashboard/users` renders the new filter bar + paginated table.
- Every filter narrows results server-side.
- Export produces a CSV reflecting the active filters.
- Biome + `tsc --noEmit` pass.

chrisgen19:
Closes #38

## Summary
- Staff Directory (`/dashboard/users`) now uses the same shadcn `Table` shell, debounced search, filter chips, dropdowns, and CSV export as `/dashboard/recognition/all`.
- `GET /api/users` gained `paginated`, `page`, `pageSize`, `search`, `roles`, `statuses`, `departmentId`, `branch`, and `export` query params. Paginated responses match the recognition shape (`pagination: { page, pageSize, total, totalPages }`).
- New client component `UserFilterBar` holds the search input, role/status chips, department/branch dropdowns, Clear, and Export actions. Role filter is only shown to SUPERADMIN (same gating as the existing role badge).
- New helper `generateUserCsv` (columns: First Name, Last Name, Email, Role, Department, Branch, Position, Status).
- Preserved existing row actions (View, Edit, Activate/Deactivate).

## Test plan
- [ ] `/dashboard/users` loads with the new filter bar + paginated table (20/page)
- [ ] Search matches across first name, last name, email, position, department
- [ ] Role chips appear only for SUPERADMIN and narrow results server-side
- [ ] Status chips (Active/Inactive) narrow results; selecting both clears the filter
- [ ] Department dropdown filters to the selected department
- [ ] Branch dropdown (ISO / Perth) filters correctly
- [ ] Clear button resets all filters
- [ ] Export CSV reflects active filters and opens correctly in a spreadsheet app
- [ ] Pagination prev/next works; page resets to 1 when any filter changes
- [ ] View / Edit / Activate-Deactivate still function
- [ ] Non-admin users are still redirected away from the page

### DIFF
--- a/app/(dashboard)/dashboard/users/_components/user-list-client.tsx
+++ b/app/(dashboard)/dashboard/users/_components/user-list-client.tsx
@@ -134,8 +134,9 @@ export function UserListClient({ currentUserRole, departments }: UserListClientP
 		setSelectedBranch("");
 	}
 
-	function buildParams(base: URLSearchParams) {
-		if (debouncedSearch) base.set("search", debouncedSearch);
+	function buildParams(base: URLSearchParams, searchOverride?: string) {
+		const effectiveSearch = searchOverride ?? debouncedSearch;
+		if (effectiveSearch) base.set("search", effectiveSearch);
 		if (selectedRoles.length > 0) base.set("roles", selectedRoles.join(","));
 		if (selectedStatuses.length > 0) base.set("statuses", selectedStatuses.join(","));
 		if (selectedDepartmentId) base.set("departmentId", selectedDepartmentId);
@@ -181,7 +182,7 @@ export function UserListClient({ currentUserRole, departments }: UserListClientP
 	async function exportUsers() {
 		setIsExporting(true);
 		try {
-			const params = buildParams(new URLSearchParams({ export: "true" }));
+			const params = buildParams(new URLSearchParams({ export: "true" }), search.trim());
 			const res = await fetch(`/api/users?${params}`);
 			if (!res.ok) throw new Error("Failed to fetch users");
 
--- a/app/(dashboard)/dashboard/users/_components/user-list-client.tsx
+++ b/app/(dashboard)/dashboard/users/_components/user-list-client.tsx
@@ -210,6 +210,12 @@ export function UserListClient({ currentUserRole, departments }: UserListClientP
 	const users = data?.data ?? [];
 	const pagination = data?.pagination;
 
+	useEffect(() => {
+		if (pagination && pagination.total > 0 && page > pagination.totalPages) {
+			setPage(pagination.totalPages);
+		}
+	}, [pagination, page]);
+
 	return (
 		<div className="max-w-7xl mx-auto space-y-6 mt-2">
 			<div className="flex flex-col gap-6 sm:flex-row sm:items-end sm:justify-between">
--- a/app/(dashboard)/dashboard/users/_components/user-list-client.tsx
+++ b/app/(dashboard)/dashboard/users/_components/user-list-client.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useQuery } from "@tanstack/react-query";
+import { useQuery, useQueryClient } from "@tanstack/react-query";
 import {
 	ChevronLeft,
 	ChevronRight,
@@ -97,6 +97,7 @@ interface UserListClientProps {
 
 export function UserListClient({ currentUserRole, departments }: UserListClientProps) {
 	const router = useRouter();
+	const queryClient = useQueryClient();
 	const showRoleFilter = currentUserRole === "SUPERADMIN";
 
 	const [page, setPage] = useState(1);
@@ -142,7 +143,7 @@ export function UserListClient({ currentUserRole, departments }: UserListClientP
 		return base;
 	}
 
-	const { data, isPending, isError, refetch } = useQuery<PaginatedResponse>({
+	const { data, isPending, isError } = useQuery<PaginatedResponse>({
 		queryKey: [
 			"users",
 			page,
@@ -171,7 +172,7 @@ export function UserListClient({ currentUserRole, departments }: UserListClientP
 		const result = await toggleUserActiveAction(userId);
 		if (result.success) {
 			toast.success("User status updated");
-			refetch();
+			queryClient.invalidateQueries({ queryKey: ["users"] });
 		} else {
 			toast.error(typeof result.error === "string" ? result.error : "Failed to update status");
 		}
--- a/app/(dashboard)/dashboard/users/_components/user-list-client.tsx
+++ b/app/(dashboard)/dashboard/users/_components/user-list-client.tsx
@@ -1,14 +1,35 @@
 "use client";
 
-import { useState } from "react";
 import { useQuery } from "@tanstack/react-query";
+import {
+	ChevronLeft,
+	ChevronRight,
+	Eye,
+	Pencil,
+	Plus,
+	Search,
+	UserCheck,
+	Users,
+	UserX,
+} from "lucide-react";
 import { useRouter } from "next/navigation";
+import { useEffect, useState } from "react";
 import { toast } from "sonner";
-import { Plus, Search, Eye, Pencil, UserX, UserCheck, Users } from "lucide-react";
-import { toggleUserActiveAction } from "@/lib/actions/user-actions";
-import { Badge } from "@/components/ui/badge";
-import { Skeleton } from "@/components/ui/skeleton";
 import { UserAvatar } from "@/components/shared/user-avatar";
+import { Badge } from "@/components/ui/badge";
+import {
+	Table,
+	TableBody,
+	TableCell,
+	TableHead,
+	TableHeader,
+	TableRow,
+} from "@/components/ui/table";
+import { toggleUserActiveAction } from "@/lib/actions/user-actions";
+import { type ExportUser, generateUserCsv } from "@/lib/users";
+import { type DepartmentOption, UserFilterBar } from "./user-filter-bar";
+
+const PAGE_SIZE = 20;
 
 interface User {
 	id: string;
@@ -25,6 +46,17 @@ interface User {
 	department: { id: string; name: string; code: string } | null;
 }
 
+interface PaginatedResponse {
+	success: boolean;
+	data: User[];
+	pagination: {
+		page: number;
+		pageSize: number;
+		total: number;
+		totalPages: number;
+	};
+}
+
 function roleBadgeVariant(role: string) {
 	switch (role) {
 		case "SUPERADMIN":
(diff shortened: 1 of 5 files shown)


app/api/users/route.ts
```diff
@@ -1,20 +1,128 @@
...
+	const search = searchParams.get("search")?.trim();
+	const roles = parseRoles(searchParams.get("roles"));
+	const statuses = parseStatuses(searchParams.get("statuses"));
+	const departmentId = searchParams.get("departmentId")?.trim();
+	const branch = parseBranch(searchParams.get("branch"));
+
+	const conditions: Prisma.UserWhereInput[] = [];
+
+	if (search) {
+		conditions.push({
```
chrisgen19: [P2] If someone types a common lookup like `Jane Doe`, none of these clauses matches because each one compares the whole string to a single column. The new server-side filter therefore stops finding users by full name, which is the most natural directory search input. Matching against a concatenated first/last expression, or an equivalent full-name field if you add one, would avoid that regression.

app/api/users/route.ts
```diff
@@ -1,20 +1,128 @@
...
+	}
+
+	const page = Math.max(1, Number(searchParams.get("page")) || 1);
+	const pageSize = Math.min(100, Math.max(1, Number(searchParams.get("pageSize")) || 20));
+
+	const [users, total] = await Promise.all([
+		prisma.user.findMany({
+			where,
+			include: { department: true },
+			orderBy: [{ firstName: "asc" }, { lastName: "asc" }],
```
chrisgen19: [P2] In the paginated path, ordering only by `firstName` and `lastName` is unstable when two staff share the same name. With `skip` and `take`, Postgres can return ties in different orders across requests, so moving between pages or refetching can duplicate a user on two pages or skip one entirely. Add a unique final sort key such as `id` to make pagination deterministic.

lib/users.ts
```diff
@@ -0,0 +1,47 @@
...
+	position: string | null;
+	branch: string | null;
+	isActive: boolean;
+	department: { name: string } | null;
+}
+
+function escapeCsvField(value: string): string {
+	const needsQuoting = value.includes(",") || value.includes('"') || value.includes("\n");
+	const escaped = needsQuoting ? `"${value.replace(/"/g, '""')}"` : value;
+	if (/^[=+\-@\t\r]/.test(escaped)) {
```
chrisgen19: [P2] If an exported value starts with a spreadsheet formula prefix and also needs quoting, for example `=SUM(1,2)`, `escaped` starts with `"` so this regex never adds the tab prefix. Excel or Sheets will still evaluate that cell as a formula, so the export remains vulnerable for exactly the quoted values that need escaping most. The prefix check needs to inspect the raw `value` before wrapping it in quotes.

app/(dashboard)/dashboard/users/_components/user-list-client.tsx
```diff
@@ -36,17 +68,103 @@ function roleBadgeVariant(role: string) {
...
+	function buildParams(base: URLSearchParams) {
+		if (debouncedSearch) base.set("search", debouncedSearch);
+		if (selectedRoles.length > 0) base.set("roles", selectedRoles.join(","));
+		if (selectedStatuses.length > 0) base.set("statuses", selectedStatuses.join(","));
+		if (selectedDepartmentId) base.set("departmentId", selectedDepartmentId);
+		if (selectedBranch) base.set("branch", selectedBranch);
+		return base;
+	}
+
+	const { data, isPending, isError, refetch } = useQuery<PaginatedResponse>({
```
chrisgen19: [P2] Now that this list is keyed by page and filters, the status-toggle path only refreshes the currently active slice. If I toggle a user on one page or filter and then revisit another cached `['users', ...]` result within the 30s stale window, React Query will serve rows and statuses from before the mutation. The toggle should invalidate the `users` query family rather than refetching only the active key.

chrisgen19 (commented):
I re-checked the current head before posting this review. Four issues still block merge:

- `app/api/users/route.ts`: search no longer matches common space-separated full-name input like `Jane Doe`, because every clause compares the whole query against a single column.
- `app/(dashboard)/dashboard/users/_components/user-list-client.tsx`: status toggles only `refetch()` the active paginated query, so other cached `users` slices can keep stale rows and statuses for up to the 30s stale window.
- `app/api/users/route.ts`: paginated ordering is not deterministic for duplicate names because it sorts only by `firstName` and `lastName` without a unique tiebreaker.
- `lib/users.ts`: CSV formula hardening checks the already-quoted string, so quoted values like `=SUM(1,2)` still bypass the tab prefix and remain spreadsheet-injectable.

Inline comments are attached on each spot.

app/api/users/route.ts
```diff
@@ -1,20 +1,128 @@
...
+	const departmentId = searchParams.get("departmentId")?.trim();
+	const branch = parseBranch(searchParams.get("branch"));
+
+	const conditions: Prisma.UserWhereInput[] = [];
+
+	if (search) {
+		conditions.push({
+			OR: [
+				{ firstName: { contains: search, mode: "insensitive" } },
+				{ lastName: { contains: search, mode: "insensitive" } },
```
chrisgen19: When the search text contains both first and last name, this query only matches if one individual column contains the entire string. For example, searching Jane Doe will not find a user whose firstName is Jane and lastName is Doe. The previous directory search matched against the concatenated full name, so moving this filter server-side regresses the most common admin lookup path.

app/api/users/route.ts
chrisgen19: This pagination order is not stable for duplicate names. If two staff members share the same firstName and lastName, their relative order can flip between requests, which means page boundaries can shift and records may be skipped or duplicated while paging through the directory. Adding a unique final tiebreaker such as id avoids that.

lib/users.ts
chrisgen19: Formula sanitization runs after the optional quoting step here. If a value starts with =, +, -, or @ and also contains a comma, quote, or newline, the escaped string begins with a quote instead of the dangerous character, so the regex no longer prefixes it and spreadsheet apps can still interpret the exported cell as a formula. The dangerous-prefix check needs to happen before wrapping the field in quotes.

app/(dashboard)/dashboard/users/_components/user-list-client.tsx
```diff
@@ -102,52 +231,79 @@ export function UserListClient({ currentUserRole }: { currentUserRole: string })
...
+				<TableSkeleton />
+			) : isError ? (
+				<div className="flex flex-col items-center justify-center rounded-xl border border-gray-200 dark:border-white/10 bg-card p-16">
+					<p className="text-[1.5rem] font-medium text-foreground">Something went wrong</p>
+					<p className="mt-2 text-base text-muted-foreground">
+						Failed to load staff. Please try again later.
+					</p>
+				</div>
+			) : users.length === 0 ? (
+				<div className="flex flex-col items-center justify-center rounded-xl border border-gray-200 dark:border-white/10 bg-card p-16">
```
chrisgen19, line 279: When an admin is on the last page of a filtered view and a mutation removes the only row on that page, the refetch still asks for the old page number. The API can then return data: [] with pagination.total still greater than zero, but this branch renders the full empty state and hides the pager, so the user is stranded on a blank No matching staff screen with no way back to page 1. Please clamp or reset page after refetch when it exceeds pagination.totalPages instead of treating every empty slice as a true empty result.

app/(dashboard)/dashboard/users/_components/user-list-client.tsx
```diff
@@ -36,53 +68,156 @@ function roleBadgeVariant(role: string) {
...
+		setDebouncedSearch("");
+		setSelectedRoles([]);
+		setSelectedStatuses([]);
+		setSelectedDepartmentId("");
+		setSelectedBranch("");
+	}
+
+	function buildParams(base: URLSearchParams) {
+		if (debouncedSearch) base.set("search", debouncedSearch);
+		if (selectedRoles.length > 0) base.set("roles", selectedRoles.join(","));
```
chrisgen19: Export reuses buildParams, which reads debouncedSearch instead of the live search input. If someone types a new query and clicks Export before the 300ms debounce fires, the downloaded CSV is generated for the previous search term even though the textbox already shows the new one. That makes the export return the wrong staff set for the visible filters.

app/api/users/route.ts
```diff
@@ -1,20 +1,131 @@
...
+
+	const where: Prisma.UserWhereInput = conditions.length > 0 ? { AND: conditions } : {};
+
+	if (isExport) {
+		const users = await prisma.user.findMany({
+			where,
+			include: { department: { select: { name: true } } },
+			orderBy: [{ firstName: "asc" }, { lastName: "asc" }, { id: "asc" }],
+			take: EXPORT_LIMIT,
+		});
```
chrisgen19, line 115: This export path silently truncates results at 10,000 rows. Once the directory grows beyond that, the query still returns success but drops every matching user after the first 10,000, so the downloaded CSV is incomplete with no warning to the admin. The endpoint needs to page through the full result set or surface the truncation explicitly to the UI.

chrisgen19:
Addressed the remaining review feedback across four commits:

**`f0ddd5b` — initial 4 issues**
- `app/api/users/route.ts`: search is now split into whitespace tokens so `Jane Doe` matches when `firstName=Jane` and `lastName=Doe`.
- `app/api/users/route.ts`: added `{ id: "asc" }` tiebreaker to paginated + export ordering for deterministic pagination.
- `lib/users.ts`: formula-injection tab prefix is applied to the raw value before quoting, so values like `=SUM(1,2)` are hardened even when they need quoting.
- `user-list-client.tsx`: status toggle now calls `queryClient.invalidateQueries({ queryKey: ["users"] })` so every cached `users` slice refreshes, not only the active page/filter.

**`d8bacb6` — orphaned page after mutation**
- `user-list-client.tsx`: added an effect that clamps `page` to `pagination.totalPages` when `total > 0` but the current page exceeds it, so admins are no longer stranded on a blank "No matching staff" screen after a filter change or status toggle.

**`d4781cf` — export debounce race**
- `user-list-client.tsx`: `buildParams` accepts an optional search override; the export path passes `search.trim()` so a CSV clicked within the 300ms debounce matches the visible filters. Paginated queries still use `debouncedSearch` to avoid a request on every keystroke.

**`967ed72` — silent 10k export truncation**
- `app/api/users/route.ts`: export response now also returns `total`, `truncated`, and `limit` via a parallel `count`.
- `user-list-client.tsx`: when `truncated` is true, the client shows a `toast.warning` naming the full total (e.g. "Exported first 10,000 of 12,345 users. Apply filters to narrow the result set.") instead of a silent success toast. Left the 10k cap in place since a staff directory should not realistically exceed that; streaming can be added later if it ever does.

`bunx tsc --noEmit` is clean after each commit.

PR #41, opened 2026-04-17
fix(users): align detail + edit pages with standard dashboard layout

### LINKED ISSUE #40: Dashboard layout consistency: user detail + edit pages deviate from standard
## Problem

`/dashboard/users/[id]` and `/dashboard/users/[id]/edit` look visually off compared to the rest of the dashboard. The broader codebase has settled on a consistent layout pattern (used by users list, departments, leaderboard, admin-settings, recognition detail, profile), but a few pages were never updated to match.

## Standard pattern used by newer pages

- **Outer wrapper:** `max-w-7xl mx-auto space-y-8 mt-2`
- **Page header:** title `text-[2.25rem] leading-tight font-medium tracking-tight`, subtitle `mt-2 text-base text-muted-foreground`, right-aligned primary action; back buttons live in a row above or beside the title block
- **Cards:** `rounded-[2rem] border border-gray-200 dark:border-white/10 bg-card shadow-[0_2px_20px_-4px_rgba(0,0,0,0.03)] overflow-hidden`
- **Card heading:** `px-8 pt-8 pb-2`, heading `text-[1.25rem] font-medium`
- **Card body:** `px-8 py-6 space-y-4`

### `/dashboard/users/[id]/page.tsx` (detail)
- Uses `max-w-5xl` instead of `max-w-7xl`.
- Back button, title, and edit button are all crammed into a single flex row — reads awkwardly on narrow screens (the title and email get squeezed between the two buttons).
- Info cards omit `overflow-hidden`, so the rounded corners can clip oddly when the inner divider bar runs edge-to-edge.
- Border uses `border-gray-200` instead of the softer `border-gray-200/60` that the rest of the app uses.

### `/dashboard/users/[id]/edit/page.tsx` (edit)
- No page-level header: the page drops straight into the form, with no title, breadcrumb, or back link — inconsistent with every other edit flow.
- `ResetPasswordForm` is wrapped in an inline `max-w-2xl` div rather than rendered as a card with the standard shell, so it visually floats at a different width than the form above it.

## Out of scope for this PR (follow-up)

These surfaced in the audit but aren't the focus right now; tracking them for later:
- `/dashboard/recognition/create/page.tsx` and `/dashboard/recognition/[id]/edit/page.tsx` use `max-w-5xl py-4` instead of the standard wrapper.
- `/dashboard/recognition/[id]/page.tsx` uses `space-y-6` and nests `max-w-4xl` on its interaction bar.
- `/dashboard/recognition/(inbox)/layout.tsx` uses `space-y-6` instead of `space-y-8`.
- `/dashboard/leaderboard/page.tsx` empty-state card uses `py-16` (acceptable for empty states but worth normalizing).

## Acceptance

- User detail page uses the standard wrapper + clean header (back link above title, edit button aligned to the right) and consistent cards.
- User edit page has a real page header (title + back link + context) and the reset-password panel follows the same card shell as the main form.
- No behavior changes — layout/styling only.

chrisgen19:
Closes #40

## Summary
- `/dashboard/users/[id]` now uses the standard `max-w-7xl` wrapper, softer `border-gray-200/60`, `overflow-hidden` cards, and surfaces Branch in Work Info. Title/email truncate cleanly so long names don't squeeze out the Edit button on narrow screens.
- `/dashboard/users/[id]/edit` gains a proper page header (back link + title + email) matching the detail page. The ad-hoc `max-w-2xl` wrapper around `ResetPasswordForm` is removed since the component already renders its own card at the correct width.
- No behavior changes — layout and styling only.

## Out of scope (tracked in #40 for follow-up)
- `/dashboard/recognition/create` + `/dashboard/recognition/[id]/edit` (`max-w-5xl py-4`)
- `/dashboard/recognition/[id]` (`space-y-6`, nested `max-w-4xl`)
- `/dashboard/recognition/(inbox)/layout.tsx` (`space-y-6`)
- `/dashboard/leaderboard` empty-state `py-16`

## Test plan
- [ ] `/dashboard/users/[id]` renders with a wider layout, consistent card styling, and Branch visible
- [ ] Long names/emails truncate and don't overlap the Edit button at small widths
- [ ] `/dashboard/users/[id]/edit` shows the new header with a working back link to the detail page
- [ ] UserForm and ResetPasswordForm sit at the same visual width, no gap/offset between them
- [ ] Dark mode renders correctly on both pages
- [ ] No regressions on Save / Reset Password actions


chrisgen19 (commented):
## Review — PR #41

### Overview
Pure layout/styling fix that brings `/dashboard/users/[id]` and `/dashboard/users/[id]/edit` in line with the dashboard's documented pattern (issue #40). No behavior change: data fetching, auth gating, and form wiring are untouched.

### What's good
- Widens detail page to `max-w-7xl`, softens border to `border-gray-200/60`, and adds `overflow-hidden` on the info cards — matches the rest of the dashboard.
- `min-w-0` + `truncate` on the header title/subtitle prevents long names and emails from squeezing the action button out of frame.
- Back link gains an `aria-label`, and section headings are promoted from `h3` → `h2` which is the correct rank under the page `h1`.
- Edit page gains a standard header so the edit flow no longer drops straight into a header-less form. Dropping the redundant `max-w-2xl` wrapper around `ResetPasswordForm` is correct — the component already renders its own card at the same width as `UserForm`.
- Surfaces the missing **Branch** field in Work Info. Adding `items-center` to the badge rows fixes a subtle vertical misalignment that was there before.
- Import ordering is cleaned up per Biome conventions.

### Findings

**1. Heading hierarchy in the edit page skips a level (minor a11y)**
- `users/[id]/edit/page.tsx` now renders an `h1` "Edit {name}".
- `user-form.tsx:119` still uses `h3` for "User Information" (and `reset-password-form.tsx:62` uses `h3` for "Reset Password").
- Accessible order should be `h1 → h2`. Since both forms are headings of equal weight under the page `h1`, the cleanest fix is to change those two component headings from `h3` to `h2`. Not a blocker, but worth a two-character follow-up.

**2. Branch displays as raw enum (`PERTH`) instead of label (`Perth`)**
- `users/[id]/page.tsx:67` — `<InfoRow label="Branch" value={user.branch} />` prints the raw Prisma enum.
- Elsewhere in the app the branch is shown using the `BRANCH_OPTIONS` label map from `_components/user-filter-bar.tsx` (`ISO` → "ISO", `PERTH` → "Perth"). Passing it through that label (or a small `formatBranch` helper) keeps capitalization consistent with the list and filter chips.

**3. Truncated values have no tooltip**
- `InfoRow` value now has `truncate`, which hides overflow silently. For long values like display names or phone numbers on a narrow column, users can't see the full value. Adding `title={value ?? undefined}` on the value span gives them a hover tooltip at zero cost.

**4. Lost divider between Role and Status (tiny)**
- The pre-change version had an `h-px` divider between Role and Status badges. The new layout drops it, which is defensible (none of the other rows have dividers), but it's a subtle visual change worth noting. Fine to keep as-is.

### Risk & test coverage
- Risk: low. Styling/layout only; no state, query, or action touched.
- Tests: none added. Project has no test runner wired up, so there's nothing to plug into — consistent with prior PRs.
- CI: `tsc --noEmit` clean per the commit; GitGuardian + Vercel should pass.

### Verdict
Ready to merge once you decide on findings 1–3. None are blockers — (2) is the most user-visible and worth grabbing before merge; (1) and (3) are fine as follow-ups.

app/(dashboard)/dashboard/users/[id]/page.tsx
```diff
@@ -1,11 +1,17 @@
...
 import { Badge } from "@/components/ui/badge";
-import { ArrowLeft, Pencil } from "lucide-react";
+import { getServerSession } from "@/lib/auth-utils";
+import { prisma } from "@/lib/db";
+import { canViewUsers } from "@/lib/permissions";
+import { BRANCH_OPTIONS } from "../_components/user-filter-bar";
+
+function formatBranch(branch: string | null): string | null {
+	if (!branch) return null;
+	return BRANCH_OPTIONS.find((b) => b.value === branch)?.label ?? branch;
```
Comment (chrisgen19) [inline app/(dashboard)/dashboard/users/[id]/page.tsx]: BRANCH_OPTIONS is being imported from a use client module, so this server page receives a client reference instead of the real array. For any user with a branch value, formatBranch() will call find() during server render and React will throw when it tries to access that client export on the server. Please move this lookup data into a shared server-safe module, or inline the branch mapping here, before using it in the page.

Issue #43, opened 2026-04-17
feat(notifications): coalesce burst reactions/comments into a single notification


chrisgen19:
## Summary

Coalesce rapid-fire card interaction notifications so a user doesn't get spammed when multiple colleagues react or comment on the same card in a short window.

## Motivation

With per-event notifications, a popular card can fire many notifications in minutes (e.g. 5 reactions → 5 bell entries). This becomes noisy and drowns out other notifications.

## Proposed behaviour

- Within a sliding window (e.g. 10 minutes), group consecutive `CARD_REACTION` or `CARD_COMMENT` notifications for the same card + same recipient.
- Message becomes \"3 people reacted to your card\" / \"2 new comments on your card\".
- When the group is unread, incoming events update the count in-place instead of creating a new row.
- Once the user marks it read (or the window expires), the next event starts a fresh notification.

## Open questions

- Window length (10 min? 1 hour?)
- Group across reaction + comment, or keep separate per type?
- Display: show the latest 2–3 actor avatars in the bell row?

## Dependencies

Depends on the base per-event notification work (see previous issue).

PR #44, opened 2026-04-17
feat(recognition): notify on reactions/comments + show reactors

### LINKED ISSUE #42: feat(recognition): notify on card reactions & comments + show reactor names
## Summary

Extend the existing notification system so users are notified when someone reacts to or comments on a recognition card, and let users see **who** reacted (not just emoji counts).

## Motivation

Today a recipient only gets a notification when the card is first received. Reactions and comments happen silently — the recipient has no way of knowing colleagues congratulated them, and the sender never finds out either. Additionally, reactions currently show counts only (e.g. \"👏 3\") with no way to see who reacted.

### 1. Notifications for reactions & comments
- New `NotificationType` enum values: `CARD_REACTION`, `CARD_COMMENT`
- Notify **both sender and recipient** (excluding the actor themselves) when:
  - Someone adds a reaction (not on remove)
  - Someone adds a comment
- One notification per event (coalescing deferred — see follow-up issue)
- Actor is never notified of their own action

### 2. Deep-link behavior on click
- Clicking a `CARD_COMMENT` notification opens the card detail page with the comment section auto-expanded and scrolled into view
- Clicking a `CARD_REACTION` notification opens the card detail page (reactions are always visible inline)

### 3. \"Who reacted\" UI
- Hover an emoji button on desktop → popover lists reactor names for that emoji
- Tap on mobile → popover toggles open
- Shows name + avatar per reactor

## Out of scope

- Notification coalescing (e.g. \"3 people reacted to your card\") — tracked separately
- Email / push notification channels

## Acceptance criteria

- [ ] Adding a reaction on a card I sent or received notifies the other participant (not me)
- [ ] Adding a comment on a card I sent or received notifies the other participant (not me)
- [ ] Removing a reaction does not create a notification
- [ ] Notification bell shows appropriate icon for the new types
- [ ] Clicking a comment notification opens the card detail with comments expanded + scrolled to
- [ ] Hover (desktop) / tap (mobile) on any emoji shows the list of users who reacted with it

chrisgen19:
Closes #42

## Summary
- Notify both sender and recipient (excluding the actor) when someone adds a reaction or comment on a recognition card. Reaction removals do not create notifications.
- Adds two `NotificationType` enum values: `CARD_REACTION`, `CARD_COMMENT`, each with an icon in the bell dropdown.
- Clicking a comment notification deep-links to `/dashboard/recognition/[cardId]?focus=comments`, which auto-expands the comment thread and scrolls it into view.
- New `ReactorPopover` shows the list of people who reacted per emoji — hover on desktop, long-press on mobile — without breaking the existing click-to-toggle reaction behaviour.

## Changes
- `prisma/schema.prisma`: extend `NotificationType` enum (`bunx prisma db push` used, no migration file since project uses db push).
- `lib/actions/interaction-actions.ts`: add `notifyCardInteraction` helper; fire notifications on reaction add and comment add.
- `app/api/recognition/[cardId]/interactions/route.ts`: include reactor `id/firstName/lastName/avatar` per emoji.
- `lib/recognition.ts`: add `CardReactionUser` type; add optional `users` to `CardReactionSummary`.
- `app/(dashboard)/dashboard/recognition/_components/reactor-popover.tsx`: new component.
- `app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx`: wrap active reactions in popover; read `?focus=comments` to auto-open + scroll.
- `components/shared/notification-bell.tsx` + `hooks/use-notifications.ts`: icons and types for the two new notification kinds; append `?focus=comments` on comment notification click.

## Test plan
- [ ] Send a card; as the recipient, add a reaction → sender sees a `CARD_REACTION` bell notification (actor does not)
- [ ] As sender, add a comment → recipient sees a `CARD_COMMENT` notification
- [ ] Remove a reaction → no notification fires
- [ ] Clicking a `CARD_COMMENT` notification lands on the detail page with the comment section expanded and scrolled into view
- [ ] Hover an emoji with reactors on desktop → popover lists their names + avatars
- [ ] Long-press an emoji on mobile → popover opens and the reaction is not toggled
- [ ] Tap an emoji on mobile → reaction toggles as before

## Follow-up
- Coalescing bursty notifications tracked in #43

### DIFF
--- a/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
@@ -41,12 +41,15 @@ export function CardInteractionBar({
 	const pendingToggles = useRef(new Set<string>());
 	const rootRef = useRef<HTMLDivElement>(null);
 
+	// Read from searchParams inside so the effect re-runs on every navigation,
+	// not just when the raw focus value changes — repeated notification clicks
+	// to the same URL must still re-open the thread.
 	useEffect(() => {
-		if (focusTarget !== "comments") return;
+		if (searchParams?.get("focus") !== "comments") return;
 		setShowComments(true);
 		setFetchEnabled(true);
 		rootRef.current?.scrollIntoView({ behavior: "smooth", block: "center" });
-	}, [focusTarget]);
+	}, [searchParams]);
 
 	const { data } = useQuery<{ success: boolean; data: CardInteractions }>({
 		queryKey: ["card-interactions", cardId, currentUserId],
--- a/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
@@ -42,9 +42,10 @@ export function CardInteractionBar({
 	const rootRef = useRef<HTMLDivElement>(null);
 
 	useEffect(() => {
-		if (focusTarget === "comments") {
-			rootRef.current?.scrollIntoView({ behavior: "smooth", block: "center" });
-		}
+		if (focusTarget !== "comments") return;
+		setShowComments(true);
+		setFetchEnabled(true);
+		rootRef.current?.scrollIntoView({ behavior: "smooth", block: "center" });
 	}, [focusTarget]);
 
 	const { data } = useQuery<{ success: boolean; data: CardInteractions }>({
@@ -174,17 +175,16 @@ export function CardInteractionBar({
 			<div className="flex flex-wrap items-center gap-1.5">
 				{/* Active reactions with counts */}
 				{activeReactions.map((r) => (
-					<ReactorPopover key={r.emoji} emoji={r.emoji} users={r.users ?? []}>
-						{(handlers) => (
+					<ReactorPopover
+						key={r.emoji}
+						emoji={r.emoji}
+						users={r.users ?? []}
+						onActivate={() => handleReaction(r.emoji)}
+					>
+						{(trigger) => (
 							<button
 								type="button"
-								onClick={() => handleReaction(r.emoji)}
-								onMouseEnter={handlers.onMouseEnter}
-								onMouseLeave={handlers.onMouseLeave}
-								onPointerDown={handlers.onPointerDown}
-								onPointerUp={handlers.onPointerUp}
-								onPointerCancel={handlers.onPointerCancel}
-								onClickCapture={handlers.onClickCapture}
+								{...trigger}
 								className={cn(
 									"inline-flex items-center gap-1 rounded-full px-2.5 py-1 text-sm transition-all select-none",
 									r.hasReacted
--- a/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
@@ -1,24 +1,27 @@
 "use client";
 
-import { useState, useRef } from "react";
-import { MessageCircle, ChevronDown, ChevronUp } from "lucide-react";
 import { useQuery, useQueryClient } from "@tanstack/react-query";
+import { ChevronDown, ChevronUp, MessageCircle } from "lucide-react";
+import { useSearchParams } from "next/navigation";
+import { useEffect, useRef, useState } from "react";
 import { toast } from "sonner";
-import { cn } from "@/lib/utils";
-import {
-	REACTION_EMOJIS,
-	type CardInteractions,
-	type CardComment,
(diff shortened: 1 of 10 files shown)


app/(dashboard)/dashboard/recognition/_components/reactor-popover.tsx
```diff
@@ -0,0 +1,124 @@
...
+			clearTimeout(longPressTimer.current);
+			longPressTimer.current = null;
+		}
+	}
+
+	function onClickCapture(e: React.MouseEvent) {
+		if (suppressNextClick.current) {
+			e.preventDefault();
+			e.stopPropagation();
+			suppressNextClick.current = false;
```
chrisgen19: On touch devices, a long press still ends up toggling the reaction. suppressNextClick is set on pointer down, but this capture handler only calls preventDefault and stopPropagation, which does not cancel the same button onClick in React or the DOM. Releasing after a long press will still reach handleReaction, so the advertised long-press to inspect without reacting flow is broken on phones and tablets.

lib/actions/interaction-actions.ts
```diff
@@ -57,14 +89,17 @@ export async function toggleReactionAction(cardId: string, emoji: string) {
 			await prisma.cardReaction.create({
 				data: { cardId, userId: session.user.id, emoji },
 			});
+			await notifyCardInteraction({
+				card,
+				actorId: session.user.id,
+				actorName: session.user.name ?? "Someone",
+				type: "CARD_REACTION",
+				emoji,
+			});
```
chrisgen19, line 103: This emits CARD_REACTION immediately after create, before the existing double-tap race has been resolved. If two toggles race while no row exists, one request can create and notify, and the other can hit P2002 and then delete the row, leaving the final state as removed while recipients still receive a reaction notification. That breaks the intended no notification on remove behavior.

app/(dashboard)/dashboard/recognition/_components/card-interaction-bar.tsx
```diff
@@ -29,11 +32,20 @@ export function CardInteractionBar({
 	initialReactions,
 }: CardInteractionBarProps) {
 	const queryClient = useQueryClient();
-	const [showComments, setShowComments] = useState(false);
+	const searchParams = useSearchParams();
+	const focusTarget = searchParams?.get("focus");
+	const [showComments, setShowComments] = useState(focusTarget === "comments");
 	// Lazy: only fetch when user first hovers (desktop) or interacts (mobile)
-	const [fetchEnabled, setFetchEnabled] = useState(false);
+	const [fetchEnabled, setFetchEnabled] = useState(focusTarget === "comments");
```
chrisgen19, line 39: focusTarget is only used to seed showComments and fetchEnabled on the first render. If the user is already on this card and the URL later changes to focus=comments, for example via another notification click or back and forward navigation, this effect scrolls the card but leaves the thread closed. The deep-link is therefore unreliable on same-page navigation.

app/(dashboard)/dashboard/recognition/_components/reactor-popover.tsx
```diff
@@ -0,0 +1,124 @@
...
+		// Desktop hover only — touch devices don't fire meaningful mouseenter before click
+		if (window.matchMedia("(hover: hover)").matches) {
+			setOpen(true);
+		}
+	}
+
+	function onMouseLeave() {
+		if (window.matchMedia("(hover: hover)").matches) {
+			setOpen(false);
+		}
```
chrisgen19: The popover closes as soon as the cursor leaves the trigger, even when the pointer is moving directly into the popover. Once the list is tall enough to need scrolling, desktop users cannot actually inspect or scroll it because the absolutely positioned panel disappears before the cursor can enter it. The hover-open state needs to stay active while the cursor is over either the trigger or the popover.

chrisgen19:
### Code review

Found 1 issue:

1. `addCommentAction` creates the comment and then calls `notifyCardInteraction` outside a transaction. If the notification insert throws, the comment is already committed but the action returns `{ success: false }` to the client — the UI reports failure, the user retries, and ends up with a duplicate comment. The sibling action `createRecognitionCardAction` in `lib/actions/recognition-actions.ts` wraps the card + notification writes in `prisma.$transaction(...)` exactly to avoid this partial-commit footgun; the same pattern should be applied here (or the notification call wrapped in a `.catch()` so it cannot fail the primary mutation).

https://github.com/chrisgen19/access-group-staff/blob/c9897e513d8e0565c54de68775b5221fbd952ce6/lib/actions/interaction-actions.ts#L148-L175

Robot Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with thumbs up. Otherwise, react with thumbs down.</sub>

PR #48, opened 2026-04-18
fix(auth): resolve redirect loop from stale session cookie (#47)

### LINKED ISSUE #47: fix(auth): redirect loop between /login and /dashboard with stale session cookie
## Problem

After running dev, users sometimes hit `ERR_TOO_MANY_REDIRECTS` on localhost.

## Root cause

Redirect loop when a stale `better-auth.session_token` cookie is present (e.g. after a DB reset or expired session):

1. Visit `/login` → `proxy.ts` sees the cookie and redirects to `/dashboard`.
2. `/dashboard` → proxy calls `auth.api.getSession()` which returns `null` (stale) → skips the `isActive` check → falls through to `next()`.
3. Dashboard layout (`app/(dashboard)/layout.tsx`) runs `getServerSession()` → `null` → redirects to `/login`.
4. Back to step 1. Loop.

The proxy was treating *cookie presence* as a proxy for *being logged in*, which disagrees with what the dashboard layout considers logged in.

## Fix

In `proxy.ts`:
- On protected routes: if a session cookie is present but `getSession` returns `null`, clear the cookie and redirect to `/login` instead of silently falling through.
- On `/login` and `/register`: verify the session is actually valid before redirecting to `/dashboard`. If stale, clear the cookie and let the page render.

## Verification

- Clear cookies, reset DB, reload `/dashboard` → lands cleanly on `/login`, no loop.
- Valid session → `/login` still redirects to `/dashboard`.
- Inactive user → still gets kicked to `/login` with cookie cleared.

chrisgen19:
## Summary
- Fixes `ERR_TOO_MANY_REDIRECTS` loop between `/login` and `/dashboard` when a stale `better-auth.session_token` cookie is present (e.g. after DB reset or session expiry).
- On protected routes, if `getSession()` returns `null` despite a cookie being present, clear the cookie and redirect to `/login` instead of silently falling through.
- On `/login` and `/register`, verify the session is actually valid before redirecting to `/dashboard`; clear the cookie if stale.

Closes #47

## Test plan
- [ ] With a stale cookie, hit `/dashboard` → lands on `/login` cleanly, no loop
- [ ] With a stale cookie, hit `/login` → page renders, cookie cleared
- [ ] Valid session → `/login` still redirects to `/dashboard`
- [ ] Inactive user → still kicked to `/login` with cookie cleared
- [ ] Logged-out user hitting `/dashboard` → redirected to `/login`

### DIFF
--- a/proxy.ts
+++ b/proxy.ts
@@ -24,20 +24,31 @@ export async function proxy(request: NextRequest) {
 		const session = await auth.api.getSession({
 			headers: request.headers,
 		});
-		if (session) {
-			const user = await prisma.user.findUnique({
-				where: { id: session.user.id },
-				select: { isActive: true },
-			});
-			if (!user?.isActive) {
-				const response = NextResponse.redirect(new URL("/login", request.url));
-				response.cookies.delete(sessionToken.name);
-				return response;
-			}
+		if (!session) {
+			const response = NextResponse.redirect(new URL("/login", request.url));
+			response.cookies.delete(sessionToken.name);
+			return response;
+		}
+		const user = await prisma.user.findUnique({
+			where: { id: session.user.id },
+			select: { isActive: true },
+		});
+		if (!user?.isActive) {
+			const response = NextResponse.redirect(new URL("/login", request.url));
+			response.cookies.delete(sessionToken.name);
+			return response;
 		}
 	}
 
 	if ((pathname === "/login" || pathname === "/register") && sessionCookie) {
+		const session = await auth.api.getSession({
+			headers: request.headers,
+		});
+		if (!session) {
+			const response = NextResponse.next();
+			response.cookies.delete(sessionToken.name);
+			return response;
+		}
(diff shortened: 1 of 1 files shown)


chrisgen19:
Addressed both review findings in f338840:

**1. `getSession` can throw on auth-page path** — extracted `resolveSession()` helper that wraps `auth.api.getSession()` in try/catch and returns `null` on failure. Applied to both the protected-route branch and the `/login`/`/register` branch, so refresh/update races and transient adapter failures now fall through to the recovery path instead of surfacing a proxy 500.

**2. `callbackUrl` open-redirect on stale-cookie auth pages** — the stale-cookie branch no longer returns `NextResponse.next()` (which preserved the original query). It now redirects to a rebuilt `/login` or `/register` URL with the `callbackUrl` run through a shared `sanitizeCallback()` helper that drops anything not starting with `/` or starting with `//`. The valid-session redirect path uses the same sanitizer for consistency.

Biome + `tsc --noEmit` pass locally.

PR #52, opened 2026-04-18
fix(auth): preserve callbackUrl on bounce + sanitize at point of use

### LINKED ISSUE #49: fix(auth): sanitize callbackUrl in login form before redirect
## Problem

`app/(auth)/login/_components/login-form.tsx:25` reads `callbackUrl` directly from `searchParams` and feeds it into:

- `router.push(callbackUrl)` after email/password sign-in (line 71)
- `signIn.social({ callbackURL })` for Google/Microsoft OAuth (lines 85, 103)

No sanitization. An unauthenticated user visiting `/login?callbackUrl=//evil.example` will be redirected to `evil.example` after a successful sign-in.

PR #48 closed this for the *stale-cookie* path by sanitizing in `proxy.ts`, but fresh (cookie-less) visits fall through to `next()` without sanitization, so the vuln is still reachable on the primary login flow.

## Fix

Sanitize `callbackUrl` at the point of use in the login form. Reuse (or extract to a shared util) the same rule `proxy.ts` uses:

- must start with `/`
- must not start with `//`
- fall back to `/dashboard` otherwise

Apply to:
- `router.push` (line 71)
- both `signIn.social` calls (lines 85, 103)

Consider also applying to the register form if it has a similar flow.

## Acceptance

- [ ] `/login?callbackUrl=//evil.example` → post-login lands on `/dashboard`, never external
- [ ] `/login?callbackUrl=https://evil.example` → post-login lands on `/dashboard`
- [ ] `/login?callbackUrl=/dashboard/users` → post-login lands on `/dashboard/users`
- [ ] Same coverage for Google and Microsoft OAuth flows
- [ ] Sanitizer is a shared util (e.g. `lib/auth/safe-callback.ts`) used by both `proxy.ts` and the form

## Related

Surfaced during review of #48.

### LINKED ISSUE #50: fix(auth): preserve callbackUrl when proxy bounces deep dashboard URLs to /login
## Problem

`proxy.ts:31–33` contains dead code:

\`\`\`ts
if (isProtected && !sessionCookie) {
  const loginUrl = new URL("/login", request.url);
  if (!pathname.startsWith("/dashboard")) {   // always false here
    loginUrl.searchParams.set("callbackUrl", pathname);
  }
  return NextResponse.redirect(loginUrl);
}
\`\`\`

`isProtected` is defined as `pathname.startsWith("/dashboard")`, so the inner guard is never true. The `callbackUrl` is never set.

Effect: an unauthenticated user hitting e.g. `/dashboard/users/42` is bounced to `/login` without a `callbackUrl`, so after login they land on `/dashboard` rather than the page they originally tried to open.

## Fix

Invert the inner guard (or remove it entirely and always set the callback on this branch):

\`\`\`ts
if (pathname !== "/dashboard") {
  loginUrl.searchParams.set("callbackUrl", pathname);
}
\`\`\`

Or simply always set it — `/dashboard` → `/dashboard` is a harmless no-op post-login.

## Important — do not ship without #49

This issue increases the number of `callbackUrl` values that reach the login form. The form currently does not sanitize `callbackUrl` before `router.push` / OAuth redirect (see #49). Fix sanitization first, or bundle the two fixes in the same PR — otherwise this change could widen the open-redirect surface.

## Acceptance

- [ ] Logged-out user visits `/dashboard/users/42` → redirected to `/login?callbackUrl=%2Fdashboard%2Fusers%2F42`
- [ ] After successful sign-in, lands on `/dashboard/users/42`
- [ ] `callbackUrl` sanitization from #49 is in place before this ships

## Related

Surfaced during review of #48. Companion issue: #49.

chrisgen19:
## Summary
- Closes #50 by inverting the dead inner guard in `proxy.ts` so `callbackUrl` is set whenever the bounced path is not `/dashboard` itself.
- Closes #49 by sanitizing `callbackUrl` at point of use in the login form so `router.push` and both OAuth `signIn.social` calls reject `//evil` and absolute URLs.
- Extracts the rule into `lib/auth/safe-callback.ts` so `proxy.ts` and the form share one definition (must start with `/`, must not start with `//`, fall back to `/dashboard`).
- Bundled because #50 widens the values reaching the form, which would widen the open-redirect surface without #49.

## Test plan
- [ ] Logged-out visit to `/dashboard/users/<id>` → redirected to `/login?callbackUrl=%2Fdashboard%2Fusers%2F<id>` → after sign-in lands on that page
- [ ] `/login?callbackUrl=//evil.example` → post-login lands on `/dashboard`, never external
- [ ] `/login?callbackUrl=https://evil.example` → post-login lands on `/dashboard`
- [ ] `/login?callbackUrl=/dashboard/users` → post-login lands on `/dashboard/users`
- [ ] Same coverage for Google and Microsoft OAuth flows

### DIFF
--- a/app/(auth)/login/_components/login-form.tsx
+++ b/app/(auth)/login/_components/login-form.tsx
@@ -8,6 +8,7 @@ import { toast } from "sonner";
 import Link from "next/link";
 import { Eye, EyeOff, Loader2 } from "lucide-react";
 import { signIn } from "@/lib/auth-client";
+import { safeCallbackOrDefault } from "@/lib/auth/safe-callback";
 import { loginSchema, type LoginInput } from "@/lib/validations/auth";
 import { AccessGroupLogo } from "@/components/shared/access-logos";
 import { GoogleIcon, MicrosoftIcon } from "@/components/shared/oauth-icons";
@@ -22,7 +23,7 @@ const OAUTH_ERROR_MESSAGES: Record<string, string> = {
 export function LoginForm() {
 	const router = useRouter();
 	const searchParams = useSearchParams();
-	const callbackUrl = searchParams.get("callbackUrl") ?? "/dashboard";
+	const callbackUrl = safeCallbackOrDefault(searchParams.get("callbackUrl"));
 	const [isLoading, setIsLoading] = useState(false);
 	const [isGoogleLoading, setIsGoogleLoading] = useState(false);
 	const [isMicrosoftLoading, setIsMicrosoftLoading] = useState(false);
--- a/lib/auth/safe-callback.ts
+++ b/lib/auth/safe-callback.ts
@@ -1,0 +1,12 @@
+export function sanitizeCallback(callbackUrl: string | null | undefined): string | null {
+	if (!callbackUrl) return null;
+	if (!callbackUrl.startsWith("/") || callbackUrl.startsWith("//")) return null;
+	return callbackUrl;
+}
+
+export function safeCallbackOrDefault(
+	callbackUrl: string | null | undefined,
+	fallback = "/dashboard",
+): string {
+	return sanitizeCallback(callbackUrl) ?? fallback;
+}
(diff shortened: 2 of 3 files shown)


chrisgen19:
Closing in favour of #55. Replaced by:
- #51 already landed the #49 sanitization fix on main with different export names (`sanitizeCallbackUrl` / `safeCallbackUrl`).
- #55 is the minimal one-line proxy.ts fix for #50, the only piece not yet on main.

Closing this PR + dependent #53 + #54 to keep the tree clean.

PR #57, opened 2026-04-18
chore(lint): resolve all biome errors and re-enable lint in CI

### LINKED ISSUE #56: Lint cleanup: resolve 43 pre-existing Biome errors and re-enable lint in CI
## Context

CI was added in #TBD but `bun run lint` failed on first run due to 43 pre-existing Biome errors and 9 warnings that had never been caught (no prior CI, no pre-commit hook). To unblock CI, the lint step was removed in 995c480. This issue tracks re-enabling it.

## Categories of violations

- `lint/style/noNonNullAssertion` — `!` operator used in several components
- `lint/complexity/useLiteralKeys` — `process.env["X"]` instead of `.X`
- `lint/correctness/useExhaustiveDependencies` — missing `useEffect` deps
- `lint/a11y/noSvgWithoutTitle` — inline SVGs missing `<title>` / `aria-label`
- `lint/a11y/useAriaPropsSupportedByRole` — invalid ARIA on `<div>`
- `assist/source/organizeImports` — imports not sorted
- Formatter — indent/wrap inconsistencies

## Plan

- [ ] Run `bunx biome check --write --unsafe .` for mechanical fixes
- [ ] Manually resolve non-null assertions (prefer `?? fallback` or narrowing)
- [ ] Add `<title>` / `aria-label` to decorative SVGs, fix invalid ARIA
- [ ] Fix `useEffect` dependency arrays
- [ ] Untrack `.claude/settings.local.json` and add to `.gitignore`
- [ ] Re-add the `Lint` step to `.github/workflows/ci.yml`
- [ ] (Optional) Add a pre-commit hook (lefthook / husky) so this doesn't recur

## Follow-up

Once this lands, consider tightening the CI to fail on warnings too (`biome check --error-on-warnings`).

chrisgen19:
Closes #56

## Summary
- All 43 pre-existing Biome lint errors resolved across 30 files
- CI now runs `bun run lint` + `bun test` (both green locally)
- `.claude/settings.local.json` untracked and gitignored (local-only config)

## Changes
- **Autofixes**: imports sorted, formatting normalized, `process.env["X"]` → `.X`
- **a11y**: loading skeletons got `role="status"`; inline SVGs got `role="img"`/`aria-label` or `aria-hidden="true"`; form labels associated via `htmlFor`/`id`
- **Type safety**: removed non-null assertions in `user-form`, `recognition-form`, `recognition-feed-widget`; replaced `as any` with a typed cast in `user-form` resolver
- **Keys**: array-index keys in skeleton loops replaced with stable string keys
- **Scoped suppressions**: one `biome-ignore` each on `components/ui/label.tsx` (primitive), `components/shared/user-avatar.tsx` (external avatar URLs), `components/ui/sidebar.tsx` (UI state cookie). Stale suppression removed from `reactor-popover.tsx`.
- **Config**: `app/globals.css` excluded from Biome (Tailwind v4 `@theme`/`@custom-variant` aren't parsed)

## Test plan
- [x] `bun run lint` — clean (0 errors, 0 warnings)
- [x] `bun test` — 17/17 pass
- [ ] CI green on this PR


app/(auth)/_components/force-light.tsx
```diff
@@ -15,7 +15,7 @@ export function ForceLight() {
 				setTheme(savedTheme.current);
 			}
 		};
-	}, []); // eslint-disable-line react-hooks/exhaustive-deps
+	}, [setTheme, theme]); // eslint-disable-line react-hooks/exhaustive-deps
```
chrisgen19: Adding `theme` here makes this effect run again immediately after `setTheme("light")` updates next-themes. On that second pass `savedTheme.current` gets overwritten with `"light"`, so cleanup restores light instead of the user's original dark/system theme. Please record the pre-login theme only once before forcing light.

app/(dashboard)/dashboard/users/_components/user-list-client.tsx
```diff
@@ -116,7 +113,7 @@ export function UserListClient({ currentUserRole, departments }: UserListClientP
 
 	useEffect(() => {
 		setPage(1);
-	}, [debouncedSearch, selectedRoles, selectedStatuses, selectedDepartmentId, selectedBranch]);
+	}, []);
```
chrisgen19: This effect also only runs on mount now. On any later page, changing search, role, status, department, or branch filters keeps the stale page number, so `/api/users` can return an empty page even when matching rows exist. Please put the filter/search state back in the dependency array so the table resets to page 1 when its inputs change.

app/(dashboard)/dashboard/recognition/_components/recognition-table.tsx
```diff
@@ -105,7 +102,7 @@ export function RecognitionTable() {
 
 	useEffect(() => {
 		setPage(1);
-	}, [debouncedSearch, selectedValues, selectedMonth, selectedYear]);
+	}, []);
```
chrisgen19: This effect now only runs on mount. If someone is already on a later page and then changes search, value, month, or year filters, the next fetch keeps the stale page number, which can skip the first matches or show an empty state even though matches exist. Please restore the filter/search dependencies so `setPage(1)` runs whenever the query inputs change.

app/(auth)/_components/force-light.tsx
```diff
@@ -5,17 +5,18 @@ import { useEffect, useRef } from "react";
...
 		setTheme("light");
 		return () => {
-			if (savedTheme.current) {
-				setTheme(savedTheme.current);
+			if (original) {
+				setTheme(original);
 			}
 		};
-	}, []); // eslint-disable-line react-hooks/exhaustive-deps
+	}, [setTheme]);
```
chrisgen19: Depending on `setTheme` here is still unsafe. In `next-themes`, that setter is recreated when the active theme changes, so this effect runs again right after `setTheme("light")`, captures `"light"` as the original value, and cleanup leaves the app stuck on light instead of restoring the user's previous theme. Please keep this mount-only or otherwise record the pre-login theme exactly once.

app/(dashboard)/dashboard/recognition/(inbox)/received/page.tsx
```diff
@@ -1,14 +1,14 @@
-import { getServerSession } from "@/lib/auth-utils";
+import { requireSession } from "@/lib/auth-utils";
 import { getUserRole, hasMinRole } from "@/lib/permissions";
 import { RecognitionFeedClient } from "../../_components/recognition-feed-client";
 
 export default async function RecognitionReceivedPage() {
-	const session = await getServerSession();
+	const session = await requireSession();
```
chrisgen19: Calling `requireSession()` just to narrow `session` introduces a new uncaught failure path here. If a user is deactivated while their session cookie is still valid, `requireSession()` throws `Account deactivated` and this page turns into a generic 500 because nothing catches that exception. The previous `getServerSession()` version relied on the dashboard layout's auth check without adding this route-level crash.

app/(dashboard)/dashboard/recognition/(inbox)/sent/page.tsx
```diff
@@ -1,14 +1,14 @@
-import { getServerSession } from "@/lib/auth-utils";
+import { requireSession } from "@/lib/auth-utils";
 import { getUserRole, hasMinRole } from "@/lib/permissions";
 import { RecognitionFeedClient } from "../../_components/recognition-feed-client";
 
 export default async function RecognitionSentPage() {
-	const session = await getServerSession();
+	const session = await requireSession();
```
chrisgen19: Same issue as `received/page.tsx`: this `requireSession()` call can throw `Account deactivated` for users whose cookie is still valid but whose account was deactivated, and nothing catches that exception here. That turns the sent-recognition route into a generic 500 even though this change was only meant to narrow the session type.

components/shared/fit-text.tsx
```diff
@@ -42,7 +42,7 @@ export function FitText({ children, className, minFontSize = 11 }: FitTextProps)
 			ro.disconnect();
 			el.style.fontSize = "";
 		};
-	}, [children, minFontSize]);
+	}, [minFontSize]);
```
chrisgen19: Dropping `children` from this effect means a mounted `FitText` instance no longer remeasures when the text itself changes without a parent resize. The `ResizeObserver` only watches the parent box, so longer replacement strings can overflow and shorter ones can stay unnecessarily shrunken. Please include the text dependency so measurement reruns on content changes.

app/(dashboard)/dashboard/recognition/(inbox)/received/page.tsx
```diff
@@ -4,11 +4,12 @@ import { RecognitionFeedClient } from "../../_components/recognition-feed-client
 
 export default async function RecognitionReceivedPage() {
 	const session = await getServerSession();
+	if (!session) return null;
```
chrisgen19: If this page is reached after the surrounding dashboard layout has already been rendered, `getServerSession()` can be `null` without the parent layout redirect running again. Returning `null` here strands the user on a blank inbox pane instead of sending them back through the login flow. Please redirect to `/login` rather than silently rendering nothing.

app/(dashboard)/dashboard/recognition/(inbox)/sent/page.tsx
```diff
@@ -4,11 +4,12 @@ import { RecognitionFeedClient } from "../../_components/recognition-feed-client
 
 export default async function RecognitionSentPage() {
 	const session = await getServerSession();
+	if (!session) return null;
```
chrisgen19: Same issue as `received/page.tsx`: when a cached dashboard layout is reused after session loss, this `return null` path leaves the sent-recognition tab blank instead of sending the user back to `/login`. Please redirect here rather than silently rendering nothing.

PR #64, opened 2026-04-18
feat(users): soft-delete users via deletedAt (closes #63)

### LINKED ISSUE #63: feat(users): soft-delete users with hidden-by-default list and restore
## Summary

Add soft-delete capability to users. Deleted users are hidden by default in `/dashboard/users`, preserve all historical data (recognitions sent/received, FK integrity), and can be viewed/restored by admins via a toggle.

## Motivation

Currently the codebase only supports deactivation (`isActive` toggle) — there is no way to remove a user. Hard deletion would break recognition history. Soft delete keeps audit trails intact while letting admins "remove" users.

### 1. Database (`prisma/schema.prisma`)
- Add `deletedAt DateTime?` to the `User` model
- Add `deletedById String?` (self-relation to the admin who deleted them) — optional but useful for audit
- Run `bunx prisma db push` in dev; migration for prod

### 2. Auth / access control
- Block soft-deleted users at the proxy/session layer (same pattern as `isActive: false`) — update `proxy.ts` and session checks in `lib/auth-utils.ts`
- Revoke existing sessions on soft delete (`prisma.session.deleteMany({ where: { userId } })`)

### 3. Server actions (`lib/actions/user-actions.ts`)
- `softDeleteUserAction(userId)` — sets `deletedAt = now()`, sets `isActive = false`, deletes sessions
- `restoreUserAction(userId)` — clears `deletedAt`, leaves `isActive` state for admin to re-enable
- Permission: ADMIN+ only; prevent self-delete; prevent deleting the last SUPERADMIN
- Return `{ success, data } | { success, error }` per project convention

### 4. Queries (default filter)
- All user list/read queries add `where: { deletedAt: null }` by default
- Files to audit: `app/api/users/route.ts`, any `prisma.user.findMany/findUnique` calls in server components, recognition user picker (`app/api/recognition/users`)
- Recognition card reads should still resolve the user (include deleted users when showing historical cards — display as \"(deleted)\" or similar)

### 5. Validation (`lib/validations/user.ts`)
- Add zod schemas for soft-delete / restore action inputs

### 6. UI (`app/(dashboard)/dashboard/users/`)
- Add **Delete** action (trash icon) next to the existing activate/deactivate toggle in `user-list-client.tsx`
- Confirmation dialog (shadcn `AlertDialog`) — warn that the user will lose access and be hidden
- Add **\"Show deleted\"** filter toggle at the top of the list
- When toggle is on: show deleted users with a muted row style + **Restore** action
- Sonner toast on success/failure

### 7. Tests
- Unit: `softDeleteUserAction`, `restoreUserAction` (permission, self-delete guard, last-superadmin guard)
- Component: confirm dialog flow, show-deleted toggle
- Skip E2E for this phase

## Out of scope
- Hard delete / purge job (can be added later as a scheduled cleanup of rows older than N days)
- Bulk delete
- Audit log table (the `deletedAt` + `deletedById` fields are enough for now)

## Acceptance criteria
- [ ] Deleted users do not appear in the default users list
- [ ] Deleted users cannot log in and existing sessions are revoked
- [ ] Admins can toggle \"Show deleted\" and restore users
- [ ] Historical recognition cards still display correctly for deleted users
- [ ] Cannot delete yourself
- [ ] Cannot delete the last SUPERADMIN
- [ ] `bun run lint` and type-check pass

chrisgen19:
Closes #63.

## Summary
- Replace the separate "deactivate" and "delete" concepts with a single soft-delete state on `User` (`deletedAt`, `deletedById`).
- Deleted users are blocked at proxy/session layer, hidden from the default staff directory, and excluded from the recognition recipient picker. Historical recognition cards still resolve their names normally.
- Admins get a trash action with an `AlertDialog` confirmation, a "Show deleted" filter toggle, and a Restore action.

## Notes
- Guards: can't self-delete; can't delete the last SUPERADMIN; existing sessions are revoked on delete.
- `?includeDeleted=true` escape hatch on `GET /api/users` for the "Show deleted" toggle.
- `department` delete-count now ignores deleted users.
- Existing dev DB: one previously-deactivated user was migrated to `deletedAt = now()` before the column drop.

## Test plan
- [ ] Log in as admin, go to `/dashboard/users`, delete a user → row disappears, toast shows success, user gets logged out.
- [ ] Toggle "Show deleted" → deleted user reappears (muted) with a Restore action.
- [ ] Restore → user returns to the default list and can log in again.
- [ ] Try to delete yourself → blocked with error toast.
- [ ] Try to delete the last SUPERADMIN → blocked with error toast.
- [ ] Send a recognition card → recipient picker does not list deleted users.
- [ ] View an old recognition card whose sender or recipient is deleted → still renders their name.

### DIFF
--- a/app/(dashboard)/dashboard/users/_components/user-list-client.tsx
+++ b/app/(dashboard)/dashboard/users/_components/user-list-client.tsx
@@ -494,7 +494,7 @@ export function UserListClient({ currentUserRole, departments }: UserListClientP
 						<AlertDialogAction
 							onClick={handleConfirmDelete}
 							disabled={isMutating}
-							className="bg-destructive text-destructive-foreground hover:bg-destructive/90"
+							className="bg-destructive text-white hover:bg-destructive/90 focus-visible:ring-destructive/50"
 						>
 							{isMutating ? "Deleting..." : "Delete"}
 						</AlertDialogAction>
--- a/app/(dashboard)/dashboard/users/_components/user-list-client.tsx
+++ b/app/(dashboard)/dashboard/users/_components/user-list-client.tsx
@@ -99,20 +99,21 @@ function TableSkeleton() {
 }
 
 interface UserListClientProps {
+	mode: "active" | "deleted";
 	currentUserRole: string;
 	departments: DepartmentOption[];
 }
 
-export function UserListClient({ currentUserRole, departments }: UserListClientProps) {
+export function UserListClient({ mode, currentUserRole, departments }: UserListClientProps) {
 	const router = useRouter();
 	const queryClient = useQueryClient();
 	const showRoleFilter = currentUserRole === "SUPERADMIN";
+	const isDeletedView = mode === "deleted";
 
 	const [page, setPage] = useState(1);
 	const [search, setSearch] = useState("");
 	const [debouncedSearch, setDebouncedSearch] = useState("");
 	const [selectedRoles, setSelectedRoles] = useState<string[]>([]);
-	const [showDeleted, setShowDeleted] = useState(false);
 	const [selectedDepartmentId, setSelectedDepartmentId] = useState("");
 	const [selectedBranch, setSelectedBranch] = useState("");
 	const [isExporting, setIsExporting] = useState(false);
@@ -127,12 +128,11 @@ export function UserListClient({ currentUserRole, departments }: UserListClientP
 	// biome-ignore lint/correctness/useExhaustiveDependencies: filter values are triggers, not read inside effect
 	useEffect(() => {
 		setPage(1);
-	}, [debouncedSearch, selectedRoles, showDeleted, selectedDepartmentId, selectedBranch]);
+	}, [debouncedSearch, selectedRoles, selectedDepartmentId, selectedBranch]);
 
 	const hasActiveFilters =
 		search.length > 0 ||
 		selectedRoles.length > 0 ||
-		showDeleted ||
 		selectedDepartmentId.length > 0 ||
 		selectedBranch.length > 0;
 
@@ -140,7 +140,6 @@ export function UserListClient({ currentUserRole, departments }: UserListClientP
 		setSearch("");
 		setDebouncedSearch("");
 		setSelectedRoles([]);
-		setShowDeleted(false);
 		setSelectedDepartmentId("");
 		setSelectedBranch("");
 	}
@@ -149,7 +148,7 @@ export function UserListClient({ currentUserRole, departments }: UserListClientP
 		const effectiveSearch = searchOverride ?? debouncedSearch;
 		if (effectiveSearch) base.set("search", effectiveSearch);
 		if (selectedRoles.length > 0) base.set("roles", selectedRoles.join(","));
-		if (showDeleted) base.set("includeDeleted", "true");
+		if (isDeletedView) base.set("onlyDeleted", "true");
 		if (selectedDepartmentId) base.set("departmentId", selectedDepartmentId);
 		if (selectedBranch) base.set("branch", selectedBranch);
 		return base;
@@ -158,10 +157,10 @@ export function UserListClient({ currentUserRole, departments }: UserListClientP
 	const { data, isPending, isError } = useQuery<PaginatedResponse>({
 		queryKey: [
 			"users",
+			mode,
 			page,
 			debouncedSearch,
 			selectedRoles,
-			showDeleted,
 			selectedDepartmentId,
 			selectedBranch,
 		],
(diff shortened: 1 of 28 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
 	"private": true,
 	"scripts": {
 		"dev": "next dev --turbopack",
-		"build": "prisma generate && prisma db push && next build",
+		"build": "prisma generate && prisma db execute --file ./scripts/pre-deploy.sql && prisma db push --accept-data-loss && next build",
```
chrisgen19: [P1] `package.json` now applies `prisma db push --accept-data-loss` before `next build` succeeds. On any deploy where the build fails after the schema change, the currently running app is left against a DB with `users.is_active` already dropped, even though the old release still queries that column, so auth/user pages can start failing immediately. The safest fix is to run this schema mutation in a separate migration/deploy phase after a successful build (or keep the old column until the new release is live).

app/(dashboard)/dashboard/users/_components/user-list-client.tsx
```diff
@@ -314,75 +327,91 @@ export function UserListClient({ currentUserRole, departments }: UserListClientP
...
+												<div className="flex justify-end gap-1 [@media(hover:hover)]:opacity-0 [@media(hover:hover)]:transition-opacity [@media(hover:hover)]:group-hover:opacity-100 focus-within:opacity-100">
+													<button
+														type="button"
+														onClick={() => router.push(`/dashboard/users/${user.id}`)}
+														className="rounded-full p-2 text-muted-foreground hover:bg-blue-50 hover:text-blue-600 dark:hover:bg-blue-500/10 transition-colors"
+														aria-label="View user"
+													>
+														<Eye size={16} />
+													</button>
+													{!isDeleted && (
```
chrisgen19, line 381: [P3] This only removes the edit button from the table row. A deleted user can still be opened with `View`, then edited or have their password reset from `/dashboard/users/[id]` and `/dashboard/users/[id]/edit`, because those routes still load deleted records. That bypasses the new restore-only deleted flow introduced in this PR; the smallest safe fix is to add the same `deletedAt` guard on the detail/edit routes (or hide the detail-page edit affordance too).

package.json
chrisgen19: [P2] The `deleted_at` backfill is only wired into the `build` script here, but the repo still exposes `db:push` and developers can still run plain `prisma db push`. In those paths `is_active` will be dropped without copying `false` rows into `deleted_at`, which silently reactivates previously deactivated users. The smallest safe fix is to move this into a real migration/shared deploy script so every schema update path preserves the old state.

lib/actions/user-actions.ts
```diff
@@ -166,19 +169,85 @@ export async function updateUserAction(userId: string, formData: unknown) {
...
+			return { success: false as const, error: "User is not deleted" };
+		}
+
 		const updated = await prisma.user.update({
 			where: { id: userId },
-			data: { isActive: !user.isActive },
+			data: { deletedAt: null, deletedById: null, isActive: true },
 		});
 
 		revalidatePath("/dashboard/users");
```
chrisgen19: [P2] When `restoreUserAction` is invoked from `/dashboard/users/deleted`, this only revalidates `/dashboard/users`. The table itself refetches because the client invalidates `["users"]`, but the tab badges/counts come from `app/(dashboard)/dashboard/users/(list)/layout.tsx`, which is not part of the revalidated path, so they stay stale until a hard refresh. Revalidating the shared layout or the deleted route as well would keep the directory state consistent.

scripts/pre-deploy.sql
```diff
@@ -0,0 +1,22 @@
...
+  IF EXISTS (
+    SELECT 1 FROM information_schema.columns
+    WHERE table_name = 'users' AND column_name = 'is_active'
+  ) THEN
+    ALTER TABLE users ADD COLUMN IF NOT EXISTS deleted_at TIMESTAMP(3);
+    ALTER TABLE users ADD COLUMN IF NOT EXISTS deleted_by_id TEXT;
+
+    -- Mirror either column into the other so old and new readers agree
+    -- during the deploy window.
+    UPDATE users SET deleted_at = NOW()
```
chrisgen19, line 17: [P1] This backfill only runs once before the deploy. In the rollout window that the comments explicitly target, an older bundle/release can still toggle a user by writing `is_active` only (for example via the old `toggleUserActiveAction`), while every new code path now authorizes and filters on `deleted_at`. In that scenario, deactivating or reactivating someone from the old release leaves `deleted_at` unchanged, so the new release keeps treating the account with the wrong status until another deploy reruns this script.

package.json
```diff
@@ -4,7 +4,7 @@
 	"private": true,
 	"scripts": {
 		"dev": "next dev --turbopack",
-		"build": "prisma generate && prisma db push && next build",
+		"build": "prisma generate && prisma db execute --file ./scripts/pre-deploy.sql && prisma db push && next build",
```
chrisgen19, line 7: [P1] The compatibility SQL only runs from `build`, so any environment that applies the schema via `bun run db:push`/`prisma db push` will add `deleted_at` without backfilling legacy `is_active = false` rows. After that, `proxy.ts`, `requireSession()`, and `/api/users` all key off `deletedAt` only, which makes previously deactivated accounts appear active again and able to authenticate until someone manually runs the SQL.

lib/auth-utils.ts
```diff
@@ -19,10 +19,10 @@ export async function requireSession() {
 	}
 	const user = await prisma.user.findUnique({
 		where: { id: session.user.id },
-		select: { isActive: true },
+		select: { deletedAt: true },
 	});
-	if (!user?.isActive) {
-		throw new Error("Account deactivated");
+	if (!user || user.deletedAt !== null) {
```
chrisgen19, line 33: [P3] `requireSession()` now throws `Account removed` for a deleted user with a still-valid session row, but `app/api/upload/avatar/route.ts` still treats only `Account deactivated` as an auth failure. In the mixed-version deploy window this patch is explicitly supporting, that makes avatar uploads return 500 instead of 401 for deleted accounts.

lib/actions/department-actions.ts
```diff
@@ -68,7 +68,9 @@ export async function updateDepartmentAction(id: string, formData: unknown) {
 export async function deleteDepartmentAction(id: string) {
 	try {
 		await requireRole("ADMIN");
-		const userCount = await prisma.user.count({ where: { departmentId: id } });
+		const userCount = await prisma.user.count({
```
chrisgen19: [P2] This now ignores soft-deleted users when deciding whether a department can be removed. Because `users.department_id` is an optional relation with `ON DELETE SET NULL`, deleting a department that is only referenced by deleted staff clears their saved department; if one of those staff members is restored later, they come back without their original department assignment. That turns a reversible soft delete into silent data loss.

lib/actions/user-actions.ts
```diff
@@ -166,19 +169,85 @@ export async function updateUserAction(userId: string, formData: unknown) {
...
+
+		if (!target) {
+			return { success: false as const, error: "User not found" };
+		}
+
+		if (target.deletedAt !== null) {
+			return { success: false as const, error: "User is already deleted" };
+		}
+
+		if (target.role === "SUPERADMIN") {
```
chrisgen19: [P2] If there are exactly two active superadmins and two delete requests arrive close together, both can observe `activeSuperadmins === 2` here and continue, because the count runs before the transactional update. In that case both soft-deletes succeed and the app ends up with zero active superadmins even though this guard is meant to prevent it; the invariant needs to be enforced inside the same transaction/conditional update.

lib/actions/user-actions.ts
```diff
@@ -166,19 +169,85 @@ export async function updateUserAction(userId: string, formData: unknown) {
...
+					// Mirror into the deprecated `isActive` column so any
+					// still-running old release renders them correctly.
+					isActive: false,
+				},
+			});
+			await tx.session.deleteMany({ where: { userId } });
+			return user;
+		});
+
+		revalidatePath("/dashboard/users", "layout");
```
chrisgen19, line 230: [P3] These delete/restore actions now revalidate the users layout, but `createUserAction` still only calls `revalidatePath("/dashboard/users")`. After moving the Active/Deleted badges into `app/(dashboard)/dashboard/users/(list)/layout.tsx`, creating a user from `/dashboard/users/new` can return to a cached layout with stale tab counts until refresh. `createUserAction` should revalidate the users layout too.

chrisgen19:
Review findings from latest pass:

1) Medium: auth consistency gap on `GET /api/users/counts`
- This route uses `auth.api.getSession()` + role check, but does not apply the deleted-user guard in `requireSession()`.
- Result: a deleted account with a still-valid/stale session token could still access this endpoint until session expiry/revocation.
- Suggestion: gate via `requireSession()` (or equivalent deletedAt check) before role validation, matching the rest of the app's auth model.

2) Test coverage gap
- I did not find automated coverage for new soft-delete/restore behavior and `onlyDeleted`/`includeDeleted` filtering paths.
- Suggest adding route/action tests for:
  - soft delete revokes access,
  - restore re-enables access,
  - `/api/users` filtering semantics for active vs deleted users.

Everything else looked solid and checks are green. Once the auth consistency issue is patched, this should be in good shape.

PR #68, opened 2026-04-20
feat(auth): sync Microsoft profile photo to avatar on first signup

### LINKED ISSUE #67: feat(auth): sync Microsoft profile photo to avatar on first signup
## Summary
When a user signs up via Microsoft OAuth at `/register` or `/login`, fetch their Microsoft profile photo from Microsoft Graph and save it to our R2 bucket as their avatar. Only runs on first signup — later manual uploads via `/dashboard/profile` are preserved.

## Background
Currently, Microsoft signups leave `users.avatar = NULL` because the OIDC `profile` claim does not include a photo URL. The photo has to be fetched from Microsoft Graph (`/me/photo/\$value`) using the OAuth access token.

## Approach
- Hook into better-auth's `databaseHooks.account.create.after` in `lib/auth.ts`.
- Only act when `providerId === \"microsoft\"` and `users.avatar` is still null (first-signup-only gate).
- Fetch the photo from Graph, upload to R2 using existing `uploadToR2` / `getAvatarKey` helpers, update `users.avatar`.
- Fire-and-forget so login redirect latency is unaffected.
- Request `User.Read` scope on the Microsoft provider.

## Regression safeguards
- Provider gate keeps Google / email-password flows untouched.
- Idempotency gate prevents overwriting a manually-uploaded avatar on re-signin.
- Fail-open: any error (no photo, 403, R2 misconfigured) is logged and swallowed — signup still succeeds.
- No schema changes; same R2 URL shape as manual uploads, so all consumers (feed, header, leaderboard) work unchanged.
- `app/api/upload/avatar/route.ts` already deletes the old R2 key on new uploads, cleanly replacing a Microsoft-sourced photo.

## Verification
- Email/password and Google signup flows unchanged.
- Microsoft signup populates avatar within a few seconds.
- Microsoft account without a photo → signup succeeds, avatar stays null.
- Manual upload after Microsoft signup → custom avatar preserved on re-signin.
- Azure app registration needs `User.Read` delegated permission granted.

chrisgen19:
Closes #67

## Summary
- Fetch Microsoft profile photo via Graph `/me/photo/$value` on first OAuth signup and store it in R2 as the user's avatar.
- First-signup-only: gated on `users.avatar === NULL`, so later manual uploads via `/dashboard/profile` are never overwritten.
- Fire-and-forget inside `databaseHooks.account.create.after` — login redirect latency is unaffected.
- Added `User.Read` scope to the Microsoft provider (required for Graph `/me/photo`).

## Regression prevention
- **Provider gate** (`providerId !== "microsoft"`) keeps Google / email-password flows untouched.
- **Idempotency gate** (`if (user.avatar) return`) ensures re-signin / account re-link never clobbers an uploaded photo.
- **Fail-open**: Graph 404 (no photo), 403 (scope not granted), R2 unconfigured, or non-image content are all silently skipped — signup still succeeds.
- **Same URL shape** as manual uploads (`getPublicUrl(getAvatarKey(...))`), so all avatar consumers (header, recognition feed, leaderboard, emails) work unchanged.
- **Existing upload route unchanged** (`app/api/upload/avatar/route.ts`). Its delete-old-key logic automatically cleans up a Microsoft-sourced R2 object if the user later uploads their own.
- **No schema / migration** — reuses `users.avatar`.
- **No token logging** — `accessToken` is used only in-memory.
- **Soft 2MB cap** on the fetched blob to avoid abusive payloads.

## Test plan
- [ ] Email/password signup: no Microsoft calls made, works as before.
- [ ] Google signup: no Microsoft calls made, works as before.
- [ ] Fresh Microsoft signup with a profile photo → avatar populated within a few seconds, renders in header.
- [ ] Microsoft signup without a profile photo → signup succeeds, `users.avatar` stays `NULL`, no user-visible error.
- [ ] Upload a custom avatar → sign out → re-signin via Microsoft → custom avatar preserved.
- [ ] Manual upload after Microsoft-synced avatar → old R2 key deleted, new one stored.
- [ ] Azure app registration has `User.Read` delegated permission granted (otherwise Graph returns 403 and hook fails gracefully).
- [x] `bun run lint` clean.

### DIFF
--- a/lib/auth.ts
+++ b/lib/auth.ts
@@ -1,5 +1,6 @@
 import { betterAuth } from "better-auth";
 import { prismaAdapter } from "better-auth/adapters/prisma";
+import { after } from "next/server";
 import { env } from "@/env";
 import { prisma } from "@/lib/db";
 import { syncMicrosoftAvatar } from "@/lib/microsoft-avatar";
@@ -50,12 +51,21 @@ export const auth = betterAuth({
 				after: async (account) => {
 					if (account.providerId !== "microsoft" || !account.accessToken) return;
 					const { userId, accessToken } = account;
-					void syncMicrosoftAvatar(userId, accessToken).catch((err) => {
-						console.error("Microsoft avatar sync failed", {
-							userId,
-							error: err instanceof Error ? err.message : String(err),
+					const run = () =>
+						syncMicrosoftAvatar(userId, accessToken).catch((err) => {
+							console.error("Microsoft avatar sync failed", {
+								userId,
+								error: err instanceof Error ? err.message : String(err),
+							});
 						});
-					});
+					// next/server `after` uses Vercel's waitUntil so the serverless
+					// invocation stays alive until the sync completes. Falls back to
+					// awaiting inline when called outside a request (e.g. seed scripts).
+					try {
+						after(run);
+					} catch {
+						await run();
+					}
 				},
 			},
 		},
--- a/lib/auth.ts
+++ b/lib/auth.ts
@@ -2,6 +2,7 @@ import { betterAuth } from "better-auth";
 import { prismaAdapter } from "better-auth/adapters/prisma";
 import { env } from "@/env";
 import { prisma } from "@/lib/db";
+import { syncMicrosoftAvatar } from "@/lib/microsoft-avatar";
 
 const socialProviders: Record<string, unknown> = {
 	google: {
@@ -20,6 +21,7 @@ if (env.MICROSOFT_CLIENT_ID && env.MICROSOFT_CLIENT_SECRET) {
 		clientId: env.MICROSOFT_CLIENT_ID,
 		clientSecret: env.MICROSOFT_CLIENT_SECRET,
 		tenantId: env.MICROSOFT_TENANT_ID ?? "common",
+		scope: ["User.Read", "openid", "profile", "email"],
 		mapProfileToUser: (profile: { given_name: string; family_name: string }) => ({
 			firstName: profile.given_name,
 			lastName: profile.family_name,
@@ -40,6 +42,22 @@ export const auth = betterAuth({
 		accountLinking: {
 			enabled: true,
 			trustedProviders: ["google", "microsoft"],
+		},
+	},
+	databaseHooks: {
+		account: {
+			create: {
+				after: async (account) => {
+					if (account.providerId !== "microsoft" || !account.accessToken) return;
+					const { userId, accessToken } = account;
+					void syncMicrosoftAvatar(userId, accessToken).catch((err) => {
+						console.error("Microsoft avatar sync failed", {
+							userId,
+							error: err instanceof Error ? err.message : String(err),
+						});
+					});
+				},
+			},
 		},
 	},
 	user: {
--- a/lib/microsoft-avatar.ts
+++ b/lib/microsoft-avatar.ts
@@ -1,5 +1,5 @@
 import { prisma } from "@/lib/db";
-import { getAvatarKey, getPublicUrl, uploadToR2 } from "@/lib/r2";
+import { deleteFromR2, getAvatarKey, getPublicUrl, uploadToR2 } from "@/lib/r2";
 
 const ALLOWED_TYPES = ["image/jpeg", "image/png", "image/webp"];
 const MAX_SIZE = 2_000_000;
@@ -27,8 +27,15 @@ export async function syncMicrosoftAvatar(userId: string, accessToken: string): 
 	const key = getAvatarKey(userId, contentType);
 	await uploadToR2(key, buffer, contentType);
 
-	await prisma.user.update({
-		where: { id: userId },
+	// Conditional update: only writes if avatar is still null. If the user
+	// uploaded a custom avatar while we were fetching from Graph, count === 0
+	// and we clean up the orphaned R2 object. Guarantees manual upload wins.
+	const result = await prisma.user.updateMany({
+		where: { id: userId, avatar: null },
 		data: { avatar: getPublicUrl(key) },
 	});
+
+	if (result.count === 0) {
+		await deleteFromR2(key).catch(() => {});
+	}
 }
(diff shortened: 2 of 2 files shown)


chrisgen19 (commented):
Code review findings:

1) High: `syncMicrosoftAvatar` has a check-then-update race that can overwrite a user’s freshly uploaded custom avatar.
- Current flow reads `user.avatar`, does async Graph fetch + R2 upload, then unconditionally updates `users.avatar`.
- If a manual upload happens while this async job is in-flight, the final update can clobber the newer user-chosen avatar.
- Suggested fix: make the final write conditional at DB level (e.g. `updateMany({ where: { id: userId, avatar: null }, data: { avatar: ... } })`) so manual uploads always win.

2) Medium: Missing regression tests for new async Microsoft avatar sync behavior.
- Please add coverage for: first Microsoft signup sets avatar, no-photo path leaves avatar null, re-signin/link does not overwrite existing avatar, and manual-upload-vs-background-sync race behavior.

Everything else looks well scoped and the fail-open handling is good.

chrisgen19:
## Review response

### ✅ High — race condition → fixed in aac0bfc

Valid finding. Replaced the check-then-update pattern with an atomic conditional update:

```ts
const result = await prisma.user.updateMany({
  where: { id: userId, avatar: null },
  data: { avatar: getPublicUrl(key) },
});

if (result.count === 0) {
  await deleteFromR2(key).catch(() => {});
}
```

This answers the open question **"should manual upload always take precedence?"** → **yes, strictly, at the DB level**. The `WHERE avatar IS NULL` guard means if a manual upload lands while Graph/R2 is in flight, our update matches zero rows and we delete the orphaned R2 object. No data loss path remains.

### ⚠️ Medium — missing regression tests → acknowledged, deferring

Valid concern, but this project currently has **zero test files** (no `vitest` / `playwright` runs wired up yet — `package.json` has a `test` script but no tests exist). Introducing the first test + harness/mocks for Prisma, `fetch`, and R2 in this PR would significantly widen its scope beyond the feature.

Proposing to track this separately: I'll open a follow-up issue to set up vitest with MSW for Graph mocking and a Prisma test-db pattern, then cover these four scenarios as the first tests in the repo:
1. First Microsoft signup sets avatar
2. No-photo (Graph 404) leaves avatar null
3. Re-login with existing avatar does not overwrite
4. Concurrent manual upload wins over background sync

Happy to block this PR on that if preferred — let me know.

chrisgen19 (commented):
Follow-up review after latest updates:

- ✅ Previous high-severity race-condition concern appears fixed.
  The new conditional DB write (`updateMany` with `where: { id, avatar: null }`) prevents Microsoft sync from overwriting a user’s manually uploaded avatar, and the orphan R2 cleanup on `count === 0` is a good addition.

- ⚠️ Remaining concern: missing automated regression tests for this new async sync behavior.
  Please add coverage (in this PR or a tracked follow-up) for:
  1) first Microsoft signup with photo sets avatar,
  2) no-photo response leaves avatar null,
  3) re-signin/account-link does not overwrite existing avatar,
  4) manual-upload-vs-background-sync concurrency where manual upload must win.

Overall this looks much safer now; test coverage is the only remaining review risk.

lib/auth.ts
```diff
@@ -42,6 +44,22 @@ export const auth = betterAuth({
 			trustedProviders: ["google", "microsoft"],
 		},
 	},
+	databaseHooks: {
+		account: {
+			create: {
+				after: async (account) => {
+					if (account.providerId !== "microsoft" || !account.accessToken) return;
+					const { userId, accessToken } = account;
+					void syncMicrosoftAvatar(userId, accessToken).catch((err) => {
```
chrisgen19: [P2] If this auth callback runs on a serverless runtime (for example Vercel/Netlify/Cloudflare), the bare `void syncMicrosoftAvatar(...)` is not tied to the request lifecycle, so the platform can tear it down as soon as sign-in completes. In that setup new Microsoft users will intermittently never get `user.avatar` populated at all. The smallest safe fix is to await the sync here, or dispatch it through Better Auth/Next's background-task mechanism instead of a detached promise.

lib/microsoft-avatar.ts
```diff
@@ -0,0 +1,41 @@
...
+
+	const rawContentType = res.headers.get("content-type") ?? "image/jpeg";
+	const contentType = rawContentType.split(";")[0]?.trim() ?? "image/jpeg";
+	if (!ALLOWED_TYPES.includes(contentType)) return;
+
+	const buffer = Buffer.from(await res.arrayBuffer());
+	if (buffer.byteLength === 0 || buffer.byteLength > MAX_SIZE) return;
+
+	const key = getAvatarKey(userId, contentType);
+	await uploadToR2(key, buffer, contentType);
```
chrisgen19, line 28: [P3] If `uploadToR2()` succeeds but `prisma.user.updateMany()` (or `getPublicUrl()`) throws, the auth hook logs the error and exits, but the just-uploaded blob is never deleted. That leaves permanent orphaned objects in R2 even though no avatar was saved. The smallest safe fix is to track `key` and delete it in a `catch`/`finally` whenever the DB update does not complete successfully.

chrisgen19:
## Review response (9f87887)

Both findings valid. Both fixed.

### ✅ P2 — serverless lifecycle

Replaced the detached `void syncMicrosoftAvatar(...)` with `next/server`'s `after()`, which hooks into Vercel's `waitUntil` so the serverless invocation stays alive until the sync promise settles. Falls back to awaiting inline if `after()` is called outside a request context (e.g. seed scripts), so the non-request path still behaves correctly.

```ts
const run = () =>
  syncMicrosoftAvatar(userId, accessToken).catch((err) => {
    console.error("Microsoft avatar sync failed", { userId, error: ... });
  });
try {
  after(run);
} catch {
  await run();
}
```

### ✅ P3 — R2 orphan on DB-write failure

Wrapped the `updateMany` in try/catch. If the DB write throws (or `getPublicUrl` throws — e.g. missing env), we delete the uploaded blob before rethrowing. The `count === 0` path (manual upload won the race) also already deletes.

```ts
try {
  const result = await prisma.user.updateMany({
    where: { id: userId, avatar: null },
    data: { avatar: getPublicUrl(key) },
  });
  if (result.count === 0) {
    await deleteFromR2(key).catch(() => {});
  }
} catch (err) {
  await deleteFromR2(key).catch(() => {});
  throw err;
}
```

Now every post-upload failure path cleans up after itself.

PR #71, opened 2026-04-20
fix(settings): invalidate setting caches across bundles via globalThis

### LINKED ISSUE #70: fix(admin): leaderboard visibility changes delay updating dashboard
## Problem

Changing **Leaderboard Visibility** under `/dashboard/admin-settings` does not immediately reflect on `/dashboard` → **Recognition Stats → Most Recognized**. Users can see up to a 30-second lag before the widget updates.

## Root cause

`lib/actions/settings-actions.ts` keeps an in-memory cache (`cachedLeaderboard`, 30s TTL) for `getLeaderboardVisibilitySettings()`. The admin Server Action calls `invalidateLeaderboardCache()` after a successful update, but in Next.js 16 (Turbopack) Server Actions and Route Handlers can be compiled into separate server bundles with distinct module-scoped variables. Result: the Server Action clears its own cache instance while the `/api/recognition/stats` Route Handler keeps serving stale data from its own copy until the TTL expires.

Impact is also worse in production serverless (Vercel) — each cold instance has its own cache, so invalidation never propagates across instances.

## Fix

Promote the cache to a true process-wide singleton via `globalThis` (same pattern already used by `lib/db/index.ts` for the Prisma client). This guarantees every bundle that imports `settings-actions.ts` reads and writes the same cache object, so invalidation immediately affects all readers.

## Scope

- `lib/actions/settings-actions.ts` — leaderboard cache only.
- OAuth and `top_recognized_limit` caches have the same shape but don't exhibit the bug in practice (admin rarely changes them mid-session); can be migrated later if needed.

## Acceptance

- Toggle visibility mode in admin settings → dashboard Most Recognized widget reflects the change immediately on next navigation/refocus, no 30s lag.

chrisgen19:
Closes #70

## Summary
- Toggling **Leaderboard Visibility** in `/dashboard/admin-settings` did not immediately reflect on `/dashboard` → Recognition Stats → Most Recognized — up to ~30s lag.
- Root cause: `lib/actions/settings-actions.ts` held its `cached*` state as plain module-scoped `let`s. Under Next.js 16 + Turbopack, Server Actions and Route Handlers can end up in separate server bundles with distinct module instances, so the Server Action's `invalidate*Cache()` cleared one copy while the `/api/recognition/stats` Route Handler kept reading its own stale copy until the TTL expired.
- Fix: promoted all three caches (leaderboard visibility, OAuth, top recognized limit) to a `globalThis`-backed store — same pattern already used for the Prisma singleton in `lib/db/index.ts`. Behavior parity is preserved (same TTL, same invalidation trigger points); only the storage location changed.

## Why the sibling caches were fixed too
- `oauthCache`: a super-admin disabling Google/Microsoft OAuth could remain accept-auth'd on `/api/auth/[...all]` for up to 30s — same bug pattern, mildly security-relevant.
- `topLimitCache`: admin changing Top-N on the same admin-settings panel hit the same 30s lag on the dashboard widget.

## Test plan
- [x] `bunx tsc --noEmit` clean
- [x] `bunx biome check lib/actions/settings-actions.ts` clean
- [x] `bun test` — 56/56 pass
- [ ] Manual: admin toggles Leaderboard Visibility → reload `/dashboard` → Most Recognized reflects new mode with no perceptible delay
- [ ] Manual: admin changes Top Recognized limit → `/dashboard` widget and `/dashboard/leaderboard` reflect the new limit on next navigation
- [ ] Manual: super-admin disables Google/Microsoft OAuth → provider immediately refused on `/api/auth/[...all]` (no 30s window)

### DIFF
--- a/lib/actions/settings-actions.ts
+++ b/lib/actions/settings-actions.ts
@@ -32,41 +32,57 @@ const TOP_RECOGNIZED_KEY = "top_recognized_limit";
 // Server Action invalidates a different cache instance than the one readers
 // (Route Handlers, Server Components) actually consult, leaking stale data for
 // up to CACHE_TTL_MS after a settings change.
+//
+// `generation` guards against a read/invalidate race: a reader captures the
+// generation before awaiting the DB, and only publishes its result back into
+// the shared cache if no invalidation happened during the await. Without this,
+// an in-flight reader holding pre-update data would re-poison the cache right
+// after an admin invalidates it.
+type CacheEntry<T> = { data: T | null; expiry: number; generation: number };
 const globalForSettingsCache = globalThis as unknown as {
-	oauthCache: { data: OAuthSettings | null; expiry: number };
-	topLimitCache: { data: number | null; expiry: number };
-	leaderboardCache: { data: LeaderboardVisibilitySettings | null; expiry: number };
+	oauthCache: CacheEntry<OAuthSettings>;
+	topLimitCache: CacheEntry<number>;
+	leaderboardCache: CacheEntry<LeaderboardVisibilitySettings>;
 };
-globalForSettingsCache.oauthCache ??= { data: null, expiry: 0 };
-globalForSettingsCache.topLimitCache ??= { data: null, expiry: 0 };
-globalForSettingsCache.leaderboardCache ??= { data: null, expiry: 0 };
+globalForSettingsCache.oauthCache ??= { data: null, expiry: 0, generation: 0 };
+globalForSettingsCache.topLimitCache ??= { data: null, expiry: 0, generation: 0 };
+globalForSettingsCache.leaderboardCache ??= { data: null, expiry: 0, generation: 0 };
 const oauthCache = globalForSettingsCache.oauthCache;
 const topLimitCache = globalForSettingsCache.topLimitCache;
 const leaderboardCache = globalForSettingsCache.leaderboardCache;
+
+function invalidate<T>(entry: CacheEntry<T>): void {
+	entry.data = null;
+	entry.expiry = 0;
+	entry.generation += 1;
+}
 
 export async function getOAuthSettings(): Promise<OAuthSettings> {
 	if (oauthCache.data && Date.now() < oauthCache.expiry) {
 		return oauthCache.data;
 	}
 
+	const generation = oauthCache.generation;
 	const settings = await prisma.appSetting.findMany({
 		where: { key: { in: [...OAUTH_KEYS] } },
 	});
 
 	const map = new Map(settings.map((s) => [s.key, s.value]));
-
-	oauthCache.data = {
+	const fresh: OAuthSettings = {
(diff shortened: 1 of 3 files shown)


lib/actions/settings-actions.ts
```diff
@@ -22,18 +22,31 @@ type OAuthKey = (typeof OAUTH_KEYS)[number];
...
+// module-scoped variables — a plain module-level `let` would mean the admin
+// Server Action invalidates a different cache instance than the one readers
+// (Route Handlers, Server Components) actually consult, leaking stale data for
+// up to CACHE_TTL_MS after a settings change.
+const globalForSettingsCache = globalThis as unknown as {
+	oauthCache: { data: OAuthSettings | null; expiry: number };
+	topLimitCache: { data: number | null; expiry: number };
+	leaderboardCache: { data: LeaderboardVisibilitySettings | null; expiry: number };
+};
+globalForSettingsCache.oauthCache ??= { data: null, expiry: 0 };
```
chrisgen19: [P2] When the cache is cold/expired and a `get*` call is already awaiting Prisma while an admin update commits, `invalidate...Cache()` only clears the shared object; the older read then resumes and writes its pre-update snapshot back into that same global cache. Because this cache is now process-wide, that stale write re-poisons every bundle in the process for another 30s (`getOAuthSettings`, `getTopRecognizedLimit`, and `getLeaderboardVisibilitySettings`). The smallest safe fix is to add a generation/version per cache and only publish read results if no invalidation happened during the await, or to remove the TTL cache for these settings.

chrisgen19: Valid, thanks. Fixed in 1c50bea: each cache entry now has a `generation` counter. Readers snapshot it before awaiting Prisma and only publish their result if no invalidation happened mid-flight; invalidation bumps the counter. The in-flight reader still returns its own value to its caller (that request predates the write), but no longer re-poisons the shared cache.

chrisgen19:
Addressed review findings:

**P2 (in d2685d9)**
- Extracted race-safe TTL cache to \`lib/cache/settings-cache.ts\` — pure logic, no Prisma/env coupling.
- \`getOrCreateGlobalEntry\` now validates the stored entry's shape and replaces it if a prior module version left an entry without \`generation\` (fixes the HMR-drift NaN trap).
- New \`lib/cache/settings-cache.test.ts\` covers cold populate, fresh hit, TTL expiry, HMR shape replacement, invalidation generation bump, and the read/invalidate race. 66/66 tests pass.
- Bonus: namespaced the \`globalThis\` keys under \`accessGroupStaff.settings.*\` instead of bare names.

**P3 — filed as follow-up issues**
- #72 — wrap globalThis keys under one parent object (cleanup).
- #73 — surface the multi-instance staleness caveat in user-facing copy or migrate to Next.js shared cache.
- #74 — coalesce concurrent cold reads behind one in-flight promise.
- #75 — playwright E2E covering admin-settings → dashboard propagation.

PR #80, opened 2026-04-20
fix(auth): handle missing Microsoft given_name/family_name claims

chrisgen19:
## Summary
- Microsoft Entra ID emits `given_name`/`family_name` only when they're configured as optional claims and populated on the user's profile. When missing, the previous mapper returned `undefined` for `firstName`/`lastName`, which violated the `required: true` additionalFields and caused OAuth sign-up to fail with `/login?error=unable_to_create_user`.
- Extracted the Google + Microsoft profile mappers to `lib/auth/profile-mappers.ts` so the logic is testable.
- Microsoft mapper now falls back to `profile.name` → `preferred_username` → email local-part, so `firstName`/`lastName` are always non-empty strings.

## Test plan
- [x] `bunx vitest run` — 89/89 pass, including 9 new cases for the mapper (regression: claims present; fallbacks: name split, multi-word last names, single-word names, `preferred_username`, email local-part, partial claims, empty profile)
- [x] `bunx tsc --noEmit` — clean
- [x] `bunx biome check` — clean
- [ ] Coworker retries Microsoft sign-up and lands in the app instead of `/login?error=unable_to_create_user`

### DIFF
--- a/lib/auth/profile-mappers.ts
+++ b/lib/auth/profile-mappers.ts
@@ -25,15 +25,25 @@ export function mapGoogleProfile(profile: GoogleProfile): MappedUser {
 	};
 }
 
+function stripEmailDomain(value: string | undefined): string | undefined {
+	if (!value) return undefined;
+	const at = value.indexOf("@");
+	return at === -1 ? value : value.slice(0, at);
+}
+
 export function mapMicrosoftProfile(profile: MicrosoftProfile): MappedUser {
-	const fallbackSource =
-		profile.name ?? profile.preferred_username ?? profile.email?.split("@")[0] ?? "";
-	const [fallbackFirst, ...fallbackRest] = fallbackSource.trim().split(/\s+/);
+	const fallbackSource = (
+		profile.name ??
+		stripEmailDomain(profile.preferred_username) ??
+		stripEmailDomain(profile.email) ??
+		""
+	).trim();
+	const [fallbackFirst, ...fallbackRest] = fallbackSource.split(/\s+/);
 	const firstName = profile.given_name ?? fallbackFirst ?? "";
 	const lastName = profile.family_name ?? (fallbackRest.join(" ") || firstName);
 	return {
 		firstName,
 		lastName,
-		name: profile.name ?? `${firstName} ${lastName}`.trim(),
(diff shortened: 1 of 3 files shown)


lib/auth/profile-mappers.ts
```diff
@@ -0,0 +1,39 @@
...
+export function mapGoogleProfile(profile: GoogleProfile): MappedUser {
+	return {
+		firstName: profile.given_name,
+		lastName: profile.family_name,
+		name: `${profile.given_name} ${profile.family_name}`,
+	};
+}
+
+export function mapMicrosoftProfile(profile: MicrosoftProfile): MappedUser {
+	const fallbackSource =
```
chrisgen19: [P2] In `lib/auth/profile-mappers.ts`, `preferred_username` is tried before `email?.split("@")[0]`, but Entra commonly returns the user's UPN/email address there. When `name` is also missing, new users are seeded with `firstName`/`lastName` like `alice@company.com`, which then shows up in avatars, profile forms, and any UI built from those fields. The smallest safe fix is to normalize `preferred_username` the same way as an email fallback, or only prefer it when it does not contain `@`.

lib/auth/profile-mappers.ts
```diff
@@ -0,0 +1,39 @@
...
+		profile.name ?? profile.preferred_username ?? profile.email?.split("@")[0] ?? "";
+	const [fallbackFirst, ...fallbackRest] = fallbackSource.trim().split(/\s+/);
+	const firstName = profile.given_name ?? fallbackFirst ?? "";
+	const lastName = profile.family_name ?? (fallbackRest.join(" ") || firstName);
+	return {
+		firstName,
+		lastName,
+		name: profile.name ?? `${firstName} ${lastName}`.trim(),
```
chrisgen19: [P2] In `lib/auth/profile-mappers.ts`, if Microsoft omits `name` and we fall back to `preferred_username` or the email local-part, this reconstructs `name` as `${firstName} ${lastName}`, so single-token fallbacks become `bob bob` / `alice@corp.com alice@corp.com`. `session.user.name` is used in notifications and headers, so these duplicated names become visible immediately after first sign-in. The smallest safe fix is to keep `name` equal to the original fallback token when `profile.name` is absent instead of rejoining the synthetic first/last pair.

PR #91, opened 2026-04-20
feat(activity-logs): audit auth events with admin viewer

### LINKED ISSUE #90: Activity logs for auth events with admin viewer
## Summary

Add an audit trail for authentication and account events, viewable by admins under Admin Settings.

## Scope (phase 1 — auth only)

- `USER_SIGNED_IN` — email or OAuth
- `USER_SIGNED_OUT`
- `SIGN_IN_FAILED` — captures attempted email + IP
- `OAUTH_ACCOUNT_LINKED` — Google / Microsoft
- `PASSWORD_CHANGED` — authenticated change
- `PASSWORD_RESET` — via reset token

## Requirements

- New `ActivityLog` table (`actorId` nullable, `action` enum, `targetType/targetId`, `metadata` JSON, `ipAddress`, `userAgent`, `createdAt`)
- Route: `/dashboard/admin-settings/activity-logs` (ADMIN+)
- Sidebar: Activity Logs as a sub-menu under Admin Settings
- Filters: actor, action, date range (from/to), target entity
- Pagination (50 per page) preserving filters
- Retention: default 180 days, configurable on `/dashboard/super-admin` (30–3650)
- Pruning: opportunistic on page load, throttled to once per 24h

## Follow-up (not in this PR)

Additional domains to log later: staff management, departments, recognition cards, interactions, admin settings, profile.

chrisgen19:
Closes #90

## Summary

- New `ActivityLog` table + `ActivityAction` enum logging 6 auth events (sign-in/out, failed sign-in, OAuth link, password change/reset).
- `/dashboard/admin-settings/activity-logs` ADMIN+ page with actor/action/date/target filters and pagination (50/page, filters preserved across pages).
- Configurable retention (default 180 days, range 30–3650) on `/dashboard/super-admin`. Pruning runs opportunistically on page load, throttled to once per 24h via an AppSetting.
- Admin Settings becomes an expandable sidebar group with Activity Logs nested underneath; auto-expands on any `/admin-settings/*` route.

## Implementation notes

- Hooks: `databaseHooks.session.create.after` catches all successful sign-ins (email + OAuth); `hooks.after` middleware handles `/sign-in/email` failures, `/sign-out`, `/change-password`, `/reset-password`; `databaseHooks.account.create.after` logs OAuth link for google/microsoft only.
- `logActivity()` is fire-and-forget (try/catch, never throws into the request path).
- Constants live in `lib/activity-log-retention.ts` so they're importable from client components without pulling in the `\"use server\"` module.

## Test plan

- [ ] Sign in with email → `USER_SIGNED_IN` entry with IP + user agent
- [ ] Sign in with wrong password → `SIGN_IN_FAILED` entry (no actor, email in metadata)
- [ ] Sign in with Google/Microsoft → both `USER_SIGNED_IN` and `OAUTH_ACCOUNT_LINKED` on first link
- [ ] Sign out → `USER_SIGNED_OUT` entry
- [ ] Change password → `PASSWORD_CHANGED` entry
- [ ] Reset password via flow → `PASSWORD_RESET` entry
- [ ] Filters work: actor, action, date range, target text
- [ ] Pagination preserves filters in URL
- [ ] STAFF user redirected away from `/dashboard/admin-settings/activity-logs`
- [ ] SUPERADMIN can change retention days on super-admin page; ADMIN cannot
- [ ] Admin Settings sidebar expands when on any `/admin-settings/*` path

### DIFF
--- a/app/(dashboard)/dashboard/admin-settings/activity-logs/page.tsx
+++ b/app/(dashboard)/dashboard/admin-settings/activity-logs/page.tsx
@@ -116,8 +116,12 @@ export default async function ActivityLogsPage({
 			<ActivityLogTable
 				logs={logs}
 				page={page}
+				pageSize={PAGE_SIZE}
 				totalPages={totalPages}
 				total={total}
+				hasActiveFilters={
+					!!params.actor || !!params.action || !!params.from || !!params.to || !!params.target
+				}
 				baseQuery={{
 					...(params.actor ? { actor: params.actor } : {}),
 					...(params.action ? { action: params.action } : {}),
--- a/app/(dashboard)/dashboard/admin-settings/activity-logs/page.tsx
+++ b/app/(dashboard)/dashboard/admin-settings/activity-logs/page.tsx
@@ -1,0 +1,131 @@
+import { redirect } from "next/navigation";
+import type { ActivityAction, Prisma } from "@/app/generated/prisma/client";
+import { pruneActivityLogsIfNeeded } from "@/lib/activity-log";
+import { requireRole } from "@/lib/auth-utils";
+import { prisma } from "@/lib/db";
+import { ActivityLogFilters } from "./_components/activity-log-filters";
+import { ActivityLogTable } from "./_components/activity-log-table";
+
+const PAGE_SIZE = 50;
+
+const ACTIONS: ActivityAction[] = [
+	"USER_SIGNED_IN",
+	"USER_SIGNED_OUT",
+	"SIGN_IN_FAILED",
+	"OAUTH_ACCOUNT_LINKED",
+	"PASSWORD_CHANGED",
+	"PASSWORD_RESET",
+];
+
+interface SearchParams {
+	actor?: string;
+	action?: string;
+	from?: string;
+	to?: string;
+	target?: string;
+	page?: string;
+}
+
+function parseDate(value: string | undefined, endOfDay = false): Date | null {
+	if (!value || !/^\d{4}-\d{2}-\d{2}$/.test(value)) return null;
+	const d = new Date(`${value}T${endOfDay ? "23:59:59.999" : "00:00:00.000"}Z`);
+	return Number.isNaN(d.getTime()) ? null : d;
+}
+
+export default async function ActivityLogsPage({
+	searchParams,
+}: {
+	searchParams: Promise<SearchParams>;
+}) {
+	try {
+		await requireRole("ADMIN");
+	} catch {
+		redirect("/dashboard");
+	}
+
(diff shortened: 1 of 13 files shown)


app/(dashboard)/dashboard/admin-settings/activity-logs/page.tsx
```diff
@@ -0,0 +1,135 @@
...
+interface SearchParams {
+	actor?: string;
+	action?: string;
+	from?: string;
+	to?: string;
+	target?: string;
+	page?: string;
+}
+
+function parseDate(value: string | undefined, endOfDay = false): Date | null {
```
chrisgen19: [P1] `parseDate()` currently appends `Z`, so a filter like `from=2026-04-20` is interpreted as midnight UTC instead of midnight in the admin's local day. In any non-UTC timezone this shifts the window by several hours and will silently exclude or include the wrong log rows near day boundaries. Please build the `Date` in local time (or normalize explicitly to the app's chosen timezone) instead of forcing UTC here.

lib/activity-log.ts
```diff
@@ -0,0 +1,77 @@
...
+		});
+	} catch (err) {
+		console.error("activity-log write failed", {
+			action: input.action,
+			error: err instanceof Error ? err.message : String(err),
+		});
+	}
+}
+
+export function extractRequestMeta(headers: Headers) {
```
chrisgen19, line 44: [P2] `extractRequestMeta()` records `x-forwarded-for` verbatim as the actor IP. In deployments where the app is reachable without a proxy that overwrites this header, any client can spoof `X-Forwarded-For` and poison the audit trail, which defeats the point of logging source IPs. Please use a trusted platform-provided IP source only, or gate `x-forwarded-for` behind an explicit trusted-proxy assumption and fall back otherwise.

app/(dashboard)/dashboard/admin-settings/activity-logs/page.tsx
```diff
@@ -0,0 +1,135 @@
...
+}) {
+	try {
+		await requireRole("ADMIN");
+	} catch {
+		redirect("/dashboard");
+	}
+
+	await pruneActivityLogsIfNeeded();
+
+	const params = await searchParams;
```
chrisgen19, line 61: [P2] This uses the raw `page` query to compute `skip` before the code knows `totalPages`. If a user opens a stale bookmark like `?page=4` after filters shrink the result set to one page, `findMany` returns no rows and the UI shows “No matching activity” even though matching entries still exist. Please count first and then clamp or redirect `page` to the valid range before running the paged query.

PR #112, opened 2026-04-21
feat(activity-logs): track authenticated page visits (USER_VISITED)

chrisgen19:
## Summary
- Adds `USER_VISITED` to the `ActivityAction` enum in `prisma/schema.prisma` and wires it through the admin activity-logs UI (filter dropdown + row label: "Visited").
- Logs a `USER_VISITED` activity entry on authenticated `/dashboard/*` visits, throttled to **once per UTC day per user** via an httpOnly `vl` cookie (`YYYY-MM-DD`, `sameSite: lax`, `maxAge: 86400`).
- Fire-and-forget (`void logActivity(...)`) — never blocks the request path.

Closes the gap where users with an active session from a prior day leave no trace in the activity log (`USER_SIGNED_IN` only fires on session creation).

## Notes

**No migration file in this PR.** The `USER_VISITED` enum value already exists in both local and production databases — it was declared inside `0002_add_missing_schema/migration.sql` (line 2) as part of the infra PR #111 that was just merged. Running `bunx prisma migrate dev` against the current schema produces no diff. A separate follow-up issue tracks a pre-existing bug in `0001_init/migration.sql` that prevents `migrate dev` from running at all; it is unrelated to this feature.

## Test plan
- [ ] Sign in, visit multiple `/dashboard/*` pages in the same day → only one `USER_VISITED` row per user
- [ ] Clear the `vl` cookie → next dashboard visit produces a new `USER_VISITED` row
- [ ] User with a valid session from a prior day visits `/dashboard` today → `USER_VISITED` logged
- [ ] `USER_VISITED` appears in the admin activity-logs filter dropdown with label "Visited"
- [ ] Existing auth events (`USER_SIGNED_IN`, `USER_SIGNED_OUT`, etc.) unaffected
- [ ] Unauthenticated visit to `/dashboard` redirects to login and logs nothing
- [x] `bun run lint` passes locally
- [ ] Vercel preview build passes
- [ ] CI (unit + E2E) passes

Closes #110

PR #114, opened 2026-04-21
chore(activity-logs): harden USER_VISITED throttle + tests

chrisgen19:
## Summary

Post-merge follow-up to #112 addressing issues found during self-review. Bundled into one small PR since all five findings touch the same files.

### Correctness
- **Scope `ag.vl` cookie value to `${userId}:${day}`** (was just `${day}`). On a shared browser, a stale cookie from a previous user no longer suppresses the new user's first visit log — the actor-scoped value forces a re-log when the signed-in user changes.
- **Rename `vl` → `ag.vl`** so it follows the existing `ag.*` cookie prefix convention (same family as `ag.session_token`), easier to grep and less prone to collisions on the same domain.

### Reliability
- **Replace `void logActivity(...)` with `after(() => logActivity(...))`** from `next/server`. On Vercel, `void` can let the instance suspend before the DB write flushes, silently dropping a small fraction of visits. `after()` keeps the invocation alive via `waitUntil` until the log resolves.

### Test hardening
- **Freeze the clock** in both `proxy.test.ts` and `lib/activity-log.test.ts` with `vi.useFakeTimers()` + `vi.setSystemTime(...)`. Previously today-derived assertions could flake if the test crossed UTC midnight between SUT invocation and the assertion.
- **New regression tests:**
  - `ag.vl` cookie scoped to a different user still triggers a log for the current user (proves the cross-user bug above is closed).
  - `NODE_ENV=production` → cookie emits the `Secure` flag (locks in the existing conditional).

## Notes

- No DB schema or migration changes.
- Cookie format change is backward-compatible: stale cookies from the previous deploy won't match any `${userId}:${day}` value, so the first post-deploy visit logs once (harmless, idempotency handled by partial unique index).
- `after()` only runs inside a request context. Tests stub it via partial `vi.mock("next/server")`.

## Test plan

- [x] `bun run lint` passes locally
- [x] `bunx vitest run` passes locally (168/168)
- [ ] CI (unit + E2E) passes
- [ ] Smoke: sign in as user A on a shared browser, sign out, sign in as user B → user B's first `/dashboard/*` hit produces a `USER_VISITED` row
- [ ] Smoke: same-day repeat visits still produce exactly one `USER_VISITED` row per user

Refs: #112

### DIFF
--- a/lib/activity-log.test.ts
+++ b/lib/activity-log.test.ts
@@ -25,16 +25,22 @@ function p2002Error(): Prisma.PrismaClientKnownRequestError {
 	);
 }
 
+const FROZEN_NOW = new Date("2026-04-21T12:34:56.000Z");
+const FROZEN_DAY_UTC = "2026-04-21T00:00:00.000Z";
+
 describe("logActivity", () => {
 	let errorSpy: ReturnType<typeof vi.spyOn>;
 
 	beforeEach(() => {
 		vi.clearAllMocks();
+		vi.useFakeTimers();
+		vi.setSystemTime(FROZEN_NOW);
 		errorSpy = vi.spyOn(console, "error").mockImplementation(() => {});
 	});
 
 	afterEach(() => {
 		errorSpy.mockRestore();
+		vi.useRealTimers();
 	});
 
 	test("USER_VISITED writes populate visitDayUtc with UTC midnight", async () => {
@@ -44,8 +50,7 @@ describe("logActivity", () => {
 
 		const arg = vi.mocked(prisma.activityLog.create).mock.calls[0][0];
 		const visitDayUtc = arg.data.visitDayUtc as Date;
-		const expected = new Date(`${new Date().toISOString().slice(0, 10)}T00:00:00.000Z`);
-		expect(visitDayUtc.toISOString()).toBe(expected.toISOString());
+		expect(visitDayUtc.toISOString()).toBe(FROZEN_DAY_UTC);
 	});
 
 	test("non-USER_VISITED writes leave visitDayUtc null", async () => {
--- a/proxy.test.ts
+++ b/proxy.test.ts
@@ -27,16 +27,6 @@ vi.mock("@/lib/auth/safe-callback", () => ({
 	sanitizeCallbackUrl: vi.fn(() => null),
 }));
 
-vi.mock("next/server", async () => {
-	const actual = await vi.importActual<typeof import("next/server")>("next/server");
-	return {
-		...actual,
-		after: (fn: () => void | Promise<void>) => {
-			void fn();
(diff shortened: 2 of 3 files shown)


chrisgen19 (commented):
Code review finding:

1) Medium — PR is not merge-ready due to deterministic E2E failure
- Affected file(s): `e2e/recognition-create.spec.ts` (failing assertion in validation test), and potentially the recognition create form validation render path.
- Why it matters: CI currently fails on this PR (`Recipient is required` never found), including retries. Even if unrelated to `proxy.ts`, this is a regression/reliability risk and blocks confidence in shipping.
- Smallest safe fix: inspect the failing Playwright trace/screenshot for this run, then either:
  - make the assertion robust to the current UI by targeting the field-level error region/accessible error linkage, or
  - fix the underlying form validation rendering so `Recipient is required` is surfaced as expected.

Notes from review:
- `proxy.ts` hardening changes (user-scoped cookie, `ag.vl`, `after(() => logActivity(...))`) look correct.
- Added unit tests in `proxy.test.ts` and `lib/activity-log.test.ts` are solid and relevant.

chrisgen19:
Pushed a revert of the `after()` change after CI showed `e2e/recognition-create.spec.ts:34` timing out at 5s (was 1.1s on #112's run).

The happy-path test (which pre-interacts with the form for ~3s before clicking) still passes; only the test that clicks immediately after `goto()` fails. Classic hydration-timing-sensitive failure. The `after()` in middleware on self-hosted `next start` (the E2E webServer) appears to change response or hydration timing enough to break that test.

Keeping:
- Cookie rename `vl` → `ag.vl`
- Cookie value scoped to `${userId}:${day}` (closes the cross-user stale-cookie bug)
- Freeze time in tests + cross-user regression test + Secure-flag test

Reverted:
- `after(() => logActivity(...))` back to `void logActivity(...)` (finding #2)

Filed #115 to investigate `after()` vs `void` properly — it's still the right pattern on Vercel, just not worth blocking this PR on.

chrisgen19 (commented):
Follow-up review:

No actionable issues found in the current head of this PR.

What I verified:
- prior E2E regression is resolved after reverting `after(...)` usage in `proxy.ts`
- CI is green (`Lint + Unit`, `E2E`, `Vercel`, `GitGuardian`)
- cookie hardening changes are correct (`ag.vl`, actor-scoped `${userId}:${utcDay}`)
- test hardening/coverage additions are solid (frozen clock + cross-user cookie + production Secure flag)

Residual risk (known, not introduced by this PR):
- `USER_VISITED` logging remains fire-and-forget (`void logActivity(...)`), so a small number of writes can still be dropped in serverless shutdown windows.

PR #116, opened 2026-04-21
fix(db): make migrations replayable on a fresh shadow DB

### LINKED ISSUE #113: fix(db): 0001_init/migration.sql has stray stdout line that breaks `prisma migrate dev`
## Problem

`prisma/migrations/0001_init/migration.sql` line 1 contains a stray line that is stdout captured from Prisma itself, not SQL:

```
Loaded Prisma config from prisma.config.ts.
```

This was introduced in PR #111 (merged as part of the migrations-workflow switch) and is present on `main` now.

## Impact

- `bunx prisma migrate dev` fails immediately with `P3006 — syntax error at or near "Loaded"` because the shadow-DB replay of `0001_init` blows up on the first line. This blocks anyone from generating a new migration locally.
- `bunx prisma migrate deploy` is **not** affected on existing DBs — both local and production already have `0001_init` marked as applied in `_prisma_migrations`, so `migrate deploy` skips it. The production deploy pipeline continues to work.
- A fresh DB (new contributor, CI ephemeral DB, disaster recovery) **would** fail to bootstrap — `migrate deploy` against an empty DB would try to apply the file and hit the syntax error.

## Verification

```
$ head -3 prisma/migrations/0001_init/migration.sql
Loaded Prisma config from prisma.config.ts.

-- CreateSchema
```

Reproducing the failure:

```
$ bunx prisma migrate dev --name anything
Migration \`0001_init\` failed to apply cleanly to the shadow database.
Error: ERROR: syntax error at or near "Loaded"
```

## Proposed fix

Remove the stray line (and the blank line immediately after it) from `prisma/migrations/0001_init/migration.sql`. The remaining SQL is valid Prisma-generated output and matches what the DB actually has.

Because `0001_init` is already recorded as applied on both local and production, the fix needs attention on the checksum:

- Option 1: update the file and run `bunx prisma migrate resolve --applied 0001_init` on each environment so the checksum in `_prisma_migrations` is refreshed. (Prisma stores a checksum and will warn on drift otherwise.)
- Option 2: update the file, leave the stored checksum alone, and accept a `prisma migrate status` drift warning until the next migration bakes in the new baseline. Safer if we are nervous about touching `_prisma_migrations` in prod.

Recommend option 1 for cleanliness, executed carefully during a low-traffic window.

## Out of scope

This is **not** blocking #110 (USER_VISITED) — that feature ships without a new migration because the enum value is already present in the DB via `0002_add_missing_schema`. See #112.

## Related

- #110 (USER_VISITED feature — surfaced this bug)
- #111 (merged PR that introduced the stray line)

chrisgen19:
## Summary

Fixes #113. Two-part fix so `prisma migrate dev` works on a fresh shadow DB:

- **`0001_init`** — drop the stray `Loaded Prisma config from prisma.config.ts.` line at the top of the SQL file (not valid SQL, was crashing replay immediately).
- **`0002_add_missing_schema`** — neutralize to a no-op. After #111 regenerated `0001_init` from the full current schema, every object `0002` originally added was already being created by `0001`, so replaying it on a shadow DB blew up with duplicate-object errors. Leaving the directory in place (with an empty body) avoids forcing a `migrate resolve --rolled-back` on every environment. The `WHY` is documented in the SQL file header.
- **`migration_lock.toml`** — check in the file Prisma auto-generates on `migrate reset`. Its own header says it belongs in version control; it was just never committed.

## Verification

- `bunx prisma migrate reset --force` — all 4 migrations apply cleanly on a fresh DB.
- `bunx prisma migrate dev --name probe --create-only` after reset — produces an empty migration (schema in sync).
- `bun run lint` — clean.
- `bunx vitest run` — 168/168 passing.

## Impact by environment

| Environment | Action required |
| --- | --- |
| **Production** | None. `migrate deploy` tolerates checksum drift on already-applied migrations. |
| **CI** | None. Unit + E2E jobs use `prisma db push`, which bypasses the migrations table. |
| **Local dev** | One-time `bunx prisma migrate reset --force` to refresh the `_prisma_migrations.checksum` rows. Seed script restores data. |

## Test plan

- [ ] Pull the branch locally, run `bunx prisma migrate reset --force`, confirm it completes without error
- [ ] Run `bunx prisma migrate dev --name probe --create-only`, confirm the generated SQL is empty (then delete the probe dir)
- [ ] CI green on the PR


chrisgen19 (commented):
Code review findings (actionable):

1) Medium — CI still does not guard migration replay correctness
- Affected files: `.github/workflows/ci.yml`, migration files under `prisma/migrations/`
- Why it matters: this PR fixes a `prisma migrate dev` replay break, but CI currently uses `prisma db push` for E2E and never replays migrations. The same class of issue can regress unnoticed until local/dev migration workflows break again.
- Smallest safe fix: add a migration-health CI step/job on Postgres that runs migration replay (e.g. `bunx prisma migrate reset --force` and/or `bunx prisma migrate deploy` on a fresh DB, plus `bunx prisma migrate dev --name ci_probe --create-only`). Fail the workflow if replay errors.

2) Low — one-time destructive local recovery is only in PR text, not repo-visible guidance
- Affected area: team migration runbook/commands (operationally tied to `prisma/migrations/*`)
- Why it matters: this PR intentionally changes checksums for already-applied migrations; local environments need a one-time `migrate reset --force`. If guidance lives only in the PR body, teammates can hit confusing migration errors later and risk accidental data loss.
- Smallest safe fix: add a short repo-visible note (or script command) documenting the one-time post-merge recovery flow and reseed expectations.

What looks good:
- Removing the stray non-SQL line from `0001_init` is correct.
- Neutralizing `0002_add_missing_schema` to avoid duplicate-object replay failures is coherent with the current migration history.

chrisgen19 (commented):
Code review finding:

1) Medium — Auto-reconciling migration checksums on every `dev` run masks migration-history integrity checks
- Affected files: `scripts/reconcile-migration-checksums.ts`, `package.json`, `README.md`
- Why it matters: Prisma checksum drift is an intentional safety signal that migration history changed after apply. Running reconciliation automatically in `predev` silently rewrites `_prisma_migrations.checksum` and can hide unintended/risky migration-history edits instead of surfacing them for explicit review.
- Smallest safe fix: remove automatic `predev` execution and keep reconciliation as an explicit/manual command only (e.g. `bun run db:reconcile-checksums`) with a narrow documented use case (PR #116-style recovery). Optionally gate execution behind an explicit env var.

What looks good:
- migration replay guard in CI is a strong improvement (`Migrations replay`)
- migration repair (`0001` cleanup + `0002` no-op) is coherent

PR #120, opened 2026-04-21
fix(activity-logs): display timestamps in Philippine Time (Asia/Manila)

### LINKED ISSUE #119: fix(activity-logs): display timestamps in Philippine Time (Asia/Manila)
## Problem

The activity log table displays timestamps using the `en-AU` locale with no explicit timezone, causing dates to render based on the server's local timezone rather than Philippine Standard Time (UTC+8).

## Solution

- Change locale from `en-AU` to `en-PH`
- Add `timeZone: "Asia/Manila"` to the `Intl.DateTimeFormat` options in `activity-log-table.tsx`

This ensures timestamps always display in PST (UTC+8) regardless of server timezone.

chrisgen19:
## Summary
- Fixes activity log timestamps displaying in server timezone instead of PST
- Changed locale from `en-AU` to `en-PH` in `activity-log-table.tsx`
- Added `timeZone: "Asia/Manila"` to `Intl.DateTimeFormat` options

Closes #119

## Test plan
- [x] Open `/dashboard/admin-settings/activity-logs`
- [x] Verify timestamps show in PST (UTC+8) — e.g. a login at 12:00 UTC should show as 8:00 PM PHT

### DIFF
--- a/app/(dashboard)/dashboard/admin-settings/activity-logs/_components/activity-log-filters.tsx
+++ b/app/(dashboard)/dashboard/admin-settings/activity-logs/_components/activity-log-filters.tsx
@@ -59,11 +59,12 @@ export function ActivityLogFilters({ users, actions, initial }: ActivityLogFilte
 			skipInitialDebounce.current = false;
 			return;
 		}
+		if (target === initial.target) return;
 		const timer = setTimeout(() => {
 			updateParam("target", target);
 		}, 300);
 		return () => clearTimeout(timer);
-	}, [target, updateParam]);
+	}, [initial.target, target, updateParam]);
 
 	const hasActiveFilters =
 		!!initial.actor || !!initial.action || !!initial.from || !!initial.to || !!initial.target;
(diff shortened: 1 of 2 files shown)


chrisgen19:
## Review (High)

**Debounced `target` filter can overwrite the URL after navigation**

**File:** `app/(dashboard)/dashboard/admin-settings/activity-logs/_components/activity-log-filters.tsx`

The debounce `useEffect` now depends on `initial.target` and returns early when `target === initial.target`.

`target` is only initialized with `useState(initial.target)`, so it does **not** update when `initial.target` changes from the server after a URL change (back/forward, new link, clearing `target` in the query). `ActivityLogFilters` is not keyed by the query string in `page.tsx`, so the instance can persist across navigations.

**Example:** If `initial.target` goes from `"hello"` to `""` but local state is still `"hello"`, the effect runs, the early return does not apply, and after 300ms `updateParam("target", "hello")` runs — putting `target` back in the URL.

Before this change, `initial.target` was not in the dependency array, so this effect did not re-run on URL-only changes (stale input was possible, but stale state was not pushed back into the URL).

**Smallest safe fix:** Sync local state when the URL-driven prop changes, e.g. a `useEffect` that does `setTarget(initial.target)` when `initial.target` changes (after sync, `target === initial.target` should make the debounce a no-op). Alternatively, remount the filter with a `key` derived from search params.

Issue #121, opened 2026-04-21
Sort dashboard users by most recent registration


chrisgen19:
## Background
The `/dashboard/users` list currently appears alphabetically in paginated mode.

## Change requested
Show users by most recent registration first so newly added staff appear at the top.

## Acceptance criteria
- `/dashboard/users` paginated list is ordered by newest users first.
- Sorting uses user creation timestamp descending (with deterministic tie-break).

PR #122, opened 2026-04-21
fix(users): sort list by newest and add joined date (#121)

chrisgen19:
## Summary
- sort `/dashboard/users` paginated results by newest registration (`createdAt desc`) so recent joins appear first
- add a `Joined` column in the users table that displays each user’s creation date
- update the users loading skeleton to keep table column alignment with the new `Joined` column

## Test plan
- [x] Open `/dashboard/users` and confirm newest registered users appear first
- [x] Verify each row shows a joined date in the new `Joined` column
- [x] Verify loading state column alignment while data is fetching

Made with [Cursor](https://cursor.com)

### DIFF
--- a/app/(dashboard)/dashboard/users/(list)/loading.tsx
+++ b/app/(dashboard)/dashboard/users/(list)/loading.tsx
@@ -49,6 +49,9 @@ export default function UsersListLoading() {
 							<TableHead>
 								<SkeletonLine className="h-3 w-12" />
 							</TableHead>
+							<TableHead>
+								<SkeletonLine className="h-3 w-12" />
+							</TableHead>
 							<TableHead className="text-right">
 								<SkeletonLine className="h-3 w-14 ml-auto" />
 							</TableHead>
@@ -77,6 +80,9 @@ export default function UsersListLoading() {
 								</TableCell>
 								<TableCell>
 									<SkeletonLine className="h-5 w-16 rounded-full" />
+								</TableCell>
+								<TableCell>
+									<SkeletonLine className="h-4 w-20" />
 								</TableCell>
 								<TableCell className="text-right">
 									<div className="flex justify-end gap-1">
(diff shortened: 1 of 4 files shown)


chrisgen19:
## Code review (findings 1–2)

### 1. Loading skeleton columns don’t match the real table (medium)

**Files:** `app/(dashboard)/dashboard/users/(list)/loading.tsx`, `app/(dashboard)/dashboard/users/_components/user-list-client.tsx` (`TableSkeleton`)

The loaded table order is **Branch → Joined → Status → Actions**, but both skeletons place the new cell **after** the Status-shaped placeholder, so **Joined and Status are swapped** in the skeleton row.

- In `loading.tsx`, the row reads: branch → pill (status) → plain line (joined); it should be branch → joined (plain) → status (pill).
- In `TableSkeleton`, the 4th cell is badge-like and the 5th is text-like; the real grid has **Joined** (text) then **Status** (badges).

**Why it matters:** This undercuts the goal of keeping loading UI aligned with the new column.

**Smallest fix:** Swap the two `TableCell` blocks (joined vs status) in both places so skeleton column order matches the data rows.

---

### 2. Export order vs on-screen list (low)

**File:** `app/api/users/route.ts`

Paginated `findMany` uses `createdAt desc`, but the **export** path still uses alphabetical `firstName` / `lastName` / `id` ordering.

**Why it matters:** Admins may assume CSV row order matches the table; it won’t.

**Smallest fix:** Either align export `orderBy` with the paginated list (`createdAt desc` + same `id` tie-breaker), or document that export is intentionally alphabetical.

chrisgen19:
Addressed the review feedback in commit `f3e8316`:

- fixed skeleton column order so loading state matches `Branch -> Joined -> Status -> Actions`
- aligned CSV export ordering with table ordering (`createdAt desc`, `id desc`)
- added `Joined` column to CSV export via `createdAt` in `lib/users.ts`

Please re-check when convenient.

PR #131, opened 2026-04-21
Backfill IT staff user metadata

### LINKED ISSUE #130: Backfill IT staff position and department for existing users
Summary
Add a one-off Prisma script to update existing staff users by email, assigning them to the IT department and filling in their job title.

Requirements
- Match users by email case-insensitively
- Set app role to STAFF
- Set department to IT
- Set position from the provided list
- Fail loudly if the IT department is missing
- Print a summary of updated and missing users
- Keep this as a manual one-off script rather than auto-running on deploy

User list
- abdul.solaiman@accessgroup.net.au -> Frontend Engineer
- chrizza.trinidad@accessgroup.net.au -> Business Intelligence Analyst
- cristian.cabacaba@accessgroup.net.au -> Lead NetSuite Technical Consultant
- cristina.delima@accessgroup.net.au -> Web Developer
- jaypee.cruz@accessgroup.net.au -> Business Intelligence Analyst
- jehnsen.enrique@accessgroup.net.au -> Integration Developer
- jeorgee.vergara@accessgroup.net.au -> NetSuite Technical Consultant
- jonathan.taruc@accessgroup.net.au -> Business Intelligence Analyst
- lester.agustin@accessgroup.net.au -> NetSuite Technical Consultant
- mil.pajares@accessgroup.net.au -> NetSuite Technical Consultant
- raymond.ricarte@accessgroup.net.au -> Business Intelligence Analyst Lead
- rob.cristobal@accessgroup.net.au -> Reports and Analytics Lead

chrisgen19:
## Summary
- add a manual one-off Prisma script to backfill existing IT staff users by email
- set app role to STAFF, set department to IT, and fill in the user position field
- add a package command for explicit execution in local or production

## Verification
- bun scripts/backfill-it-staff-users.ts
- bunx @biomejs/biome check scripts/backfill-it-staff-users.ts package.json

Closes #130


chrisgen19:
The change adds a manual one-off script at `scripts/backfill-it-staff-users.ts` and a package command in `package.json`. It matches existing users by email case-insensitively, sets `position`, forces app `role` to `STAFF`, assigns the `IT` department, and aborts `--apply` if any listed user is missing.

It will not auto-run on deploy as written, and that is the safer choice. Right now it only runs if someone explicitly executes:

```bash
bun run db:backfill:it-staff
```

For production, that means after the PR is merged and deployed, you or I would run that command once against the production `DATABASE_URL`. If you wired it into startup/build/deploy, it would run every deploy, which is usually the wrong behavior for a one-time data backfill.

I also dry-ran it against your current local DB. It matched 9 users and reported 3 missing emails:

- `chrizza.trinidad@accessgroup.net.au`
- `raymond.ricarte@accessgroup.net.au`
- `rob.cristobal@accessgroup.net.au`

If you want, I can next either update the script list to match the exact emails in production, or run the local/apply version once you confirm those 3 addresses.

chrisgen19:
Follow-up change pushed: the PR now supports a guarded production auto-run path.

How it works:
- Next 16 `instrumentation.ts` checks `RUN_IT_STAFF_BACKFILL`
- if `RUN_IT_STAFF_BACKFILL=true`, the server runs the IT staff backfill during Node startup
- it uses a Postgres advisory lock so multiple instances do not all run it concurrently
- it writes an `app_settings` completion marker after success, so later boots self-skip even if the flag is left on
- the manual commands still exist:
  - `bun run db:backfill:it-staff:dry-run`
  - `bun run db:backfill:it-staff`

Operationally, this means you can turn the flag on for one production deploy, verify success, then turn the flag off and remove the backfill code in a cleanup PR.

chrisgen19:
## Review findings

### High
- **Startup backfill can crash the server on recoverable data issues**
  - **Affected files:** `instrumentation-node.ts`, `lib/backfills/it-staff-users.ts`
  - **Why it matters:** `instrumentation-node.ts` calls `runItStaffUserBackfillOnceOnStartup()` inside a `void` async flow without a `catch`. If the backfill throws (for example, one email missing in apply mode), the promise rejection can terminate the Node process during boot. That turns a one-off data issue into an availability risk.
  - **Smallest safe fix:** Wrap the startup invocation in `try/catch` and swallow after structured logging (or gate with a "fail-open vs fail-closed" env flag). Catch error, log it, skip marker write, continue app startup.

### Medium
- **Apply mode is non-atomic and can leave partial user updates**
  - **Affected file:** `lib/backfills/it-staff-users.ts`
  - **Why it matters:** Per-user updates are executed one-by-one outside a transaction. If an update fails mid-loop (connection blip, constraint issue), earlier users stay changed while later users are untouched.
  - **Smallest safe fix:** Run the apply path in a single `prisma.$transaction(...)` (or grouped bulk updates where possible) so updates are all-or-nothing.

### Low
- **Missing direct tests for startup guard and idempotency paths**
  - **Affected area:** `lib/backfills/it-staff-users.ts`, `instrumentation-node.ts`
  - **Why it matters:** Logic now includes advisory lock behavior, marker checks, and startup execution path. CI is green, but there is no focused unit/integration coverage for: already-completed marker, lock unavailable, missing IT department, and failure behavior during startup invocation.
  - **Smallest safe fix:** Add unit tests with Prisma mocked for:
    - marker exists => skip
    - lock unavailable => skip
    - apply success => marker upsert
    - apply failure => no marker, error logged/handled at startup layer

### Summary
- Strong implementation overall (lock + marker + manual scripts are good operational controls).
- Not merge-blocked for data correctness alone, but the unhandled startup failure path should be fixed before merge if this will run in production boot flow.

PR #142, opened 2026-04-23
fix: prevent sidebar nav flicker on reload

### LINKED ISSUE #141: Fix sidebar nav links flickering on reload
## Summary
Admin and super-admin sidebar links briefly disappear on hard reload before reappearing.

## Cause
SidebarNav is a client component and currently falls back to "STAFF" before useSession() hydrates, so role-gated links are filtered out on the first render.

## Fix
Pass the server-known user role from the dashboard layout into the header/sidebar mobile trigger and use that as the initial fallback until the client session resolves.

## Acceptance Criteria
- Admin and super-admin links do not disappear briefly on hard reload.
- Desktop and mobile sidebar paths use the same initial role.
- Existing sidebar behavior remains unchanged after hydration.

chrisgen19:
## Summary
- pass the server-known user role from the dashboard layout into the header and sidebar
- use that role as the initial fallback before useSession() hydrates
- keep desktop and mobile sidebar role gating consistent on first render

## Testing
- bun x tsc --noEmit

Closes #141

### DIFF
--- a/app/(dashboard)/layout.tsx
+++ b/app/(dashboard)/layout.tsx
@@ -10,12 +10,13 @@ export default async function DashboardLayout({ children }: { children: React.Re
 	if (!session) redirect("/login");
 
 	const helpMeEnabled = await getHelpMeEnabled();
+	const initialUserRole = session.user.role as "STAFF" | "ADMIN" | "SUPERADMIN";
 
 	return (
 		<div className="flex min-h-screen bg-background">
-			<DashboardSidebar helpMeEnabled={helpMeEnabled} />
+			<DashboardSidebar helpMeEnabled={helpMeEnabled} initialUserRole={initialUserRole} />
 			<div className="flex flex-1 flex-col bg-card sm:my-2 sm:mr-2 sm:rounded-l-[2rem] shadow-sm border border-gray-100 dark:border-white/5">
-				<DashboardHeader helpMeEnabled={helpMeEnabled} />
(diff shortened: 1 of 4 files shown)


chrisgen19 (commented):
## Review findings

### High
- **Potential privilege UI regression in role fallback** (`components/shared/dashboard-sidebar.tsx`, `app/(dashboard)/layout.tsx`)
  - `SidebarNav` now falls back to `initialUserRole` whenever `useSession()` has no user. That solves hydration flicker, but it also applies later (for example after client session expiry), so privileged nav can remain visible longer than intended.
  - **Smallest safe fix:** use `initialUserRole` only while session is pending/hydrating; after hydration, if role is missing, fall back to `STAFF` (or redirect).

### Medium
- **Unrelated layout-width change may cause visual regressions** (`components/shared/dashboard-sidebar.tsx`)
  - Sidebar width changed from `w-[13.5rem]` to `w-72` in a PR scoped to reload flicker.
  - **Smallest safe fix:** revert to prior width unless this is intentional and validated.

### Low
- **Missing tests for new role-fallback behavior** (`components/shared/dashboard-sidebar.test.tsx`)
  - No targeted tests for hydration fallback vs post-hydration missing session behavior.
  - **Smallest safe fix:** add tests for (1) pending session uses server role, (2) missing role after hydration downgrades safely, and (3) desktop/mobile parity.

PR #146, opened 2026-04-23
Fix dashboard recognition feed scrolling

### LINKED ISSUE #145: Use native page scroll for dashboard recognition feed
Update /dashboard so the recognition feed uses the browser's native page scroll instead of an internal scroll panel. Make Recognition Stats sticky on desktop, preserve mobile ordering, and keep loading UI consistent.

chrisgen19:
## Summary
- let the dashboard recognition feed use native browser/page scrolling
- keep Recognition Stats visible with a measured sticky offset after the widget bottom enters the viewport
- align the loading skeleton with the updated dashboard layout

## Verification
- bun run lint

Closes #145

### DIFF
--- a/app/(dashboard)/dashboard/_components/recognition-feed-widget.tsx
+++ b/app/(dashboard)/dashboard/_components/recognition-feed-widget.tsx
@@ -62,7 +62,7 @@ export function RecognitionFeedWidget({ currentUserId, isAdmin }: RecognitionFee
 					))}
 				</div>
 			</div>
-			<div className="px-6 pb-6 overflow-y-auto max-h-[calc(100vh-12rem)]" role="tabpanel">
+			<div className="px-6 pb-6" role="tabpanel">
 				<RecognitionFeed
 					filter={activeTab}
 					showTitle={false}
--- a/app/(dashboard)/dashboard/_components/stats-widget.tsx
+++ b/app/(dashboard)/dashboard/_components/stats-widget.tsx
@@ -2,7 +2,7 @@
 
 import { useQuery, useQueryClient } from "@tanstack/react-query";
 import { Calendar, Heart, Inbox, Lock, Medal, Send, Trophy } from "lucide-react";
-import { useEffect, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { SkeletonCard, SkeletonLine } from "@/components/shared/skeleton-primitives";
 import { UserAvatar } from "@/components/shared/user-avatar";
 import { Tooltip, TooltipContent, TooltipTrigger } from "@/components/ui/tooltip";
@@ -207,7 +207,7 @@ function StatItem({
 function StatsWidgetSkeleton() {
 	return (
 		<SkeletonCard
-			className="p-6 animate-pulse flex flex-col gap-6 h-full"
+			className="p-6 animate-pulse flex flex-col gap-6"
 			role="status"
 			aria-busy="true"
 			aria-label="Loading recognition stats"
@@ -328,7 +328,7 @@ export function StatsWidget() {
 
 	if (isError || !data?.data) {
 		return (
-			<div className="rounded-[2rem] border border-gray-200/60 dark:border-white/10 bg-card p-6 shadow-[0_2px_20px_-4px_rgba(0,0,0,0.03)] h-full">
+			<div className="rounded-[2rem] border border-gray-200/60 dark:border-white/10 bg-card p-6 shadow-[0_2px_20px_-4px_rgba(0,0,0,0.03)]">
 				<h3 className="text-[1.25rem] font-medium text-foreground tracking-tight mb-2">
 					Recognition Stats
 				</h3>
@@ -346,7 +346,7 @@ export function StatsWidget() {
 	const showEmpty = resolvedVisibility.visible && stats.topRecipients.length === 0;
 
 	return (
-		<div className="rounded-[2rem] border border-gray-200/60 dark:border-white/10 bg-card p-6 shadow-[0_2px_20px_-4px_rgba(0,0,0,0.03)] flex flex-col gap-6 h-full">
+		<div className="rounded-[2rem] border border-gray-200/60 dark:border-white/10 bg-card p-6 shadow-[0_2px_20px_-4px_rgba(0,0,0,0.03)] flex flex-col gap-6">
 			<h3 className="text-[1.25rem] font-medium text-foreground tracking-tight shrink-0">
 				Recognition Stats
 			</h3>
@@ -470,6 +470,47 @@ export function StatsWidget() {
 					</div>
 				</div>
 			) : null}
+		</div>
+	);
+}
+
+export function StickyStatsWidget() {
+	const widgetRef = useRef<HTMLDivElement>(null);
+	const [stickyTop, setStickyTop] = useState(32);
+
+	useEffect(() => {
+		const widget = widgetRef.current;
+		if (!widget) return;
+
+		const updateStickyOffset = () => {
+			const viewportHeight = window.innerHeight;
+			const widgetHeight = widget.offsetHeight;
+			const offset = 32;
+			const nextTop =
+				widgetHeight + offset * 2 > viewportHeight
+					? viewportHeight - widgetHeight - offset
+					: offset;
+
+			setStickyTop(nextTop);
+		};
+
+		updateStickyOffset();
+
+		const resizeObserver =
+			typeof ResizeObserver === "undefined" ? null : new ResizeObserver(updateStickyOffset);
+
+		resizeObserver?.observe(widget);
+		window.addEventListener("resize", updateStickyOffset);
+
+		return () => {
+			resizeObserver?.disconnect();
+			window.removeEventListener("resize", updateStickyOffset);
+		};
+	}, []);
+
+	return (
+		<div ref={widgetRef} className="lg:sticky" style={{ top: stickyTop }}>
+			<StatsWidget />
 		</div>
 	);
 }
--- a/app/(dashboard)/dashboard/_components/stats-widget.tsx
+++ b/app/(dashboard)/dashboard/_components/stats-widget.tsx
@@ -463,7 +463,7 @@ export function StatsWidget() {
 						<Trophy size={16} className="text-primary" />
 						<h4 className="text-sm font-medium text-foreground/70">Most Recognized</h4>
 					</div>
-					<div className="flex items-center justify-center rounded-2xl border border-dashed border-gray-200 dark:border-white/10 bg-muted/30 px-6 py-8 text-center">
+					<div className="flex min-h-48 items-center justify-center rounded-2xl border border-dashed border-gray-200 dark:border-white/10 bg-muted/30 px-6 py-8 text-center">
 						<p className="text-sm text-muted-foreground">
 							No recognitions yet this month — be the first!
 						</p>
@@ -483,7 +483,7 @@ export function StickyStatsWidget() {
 		if (!widget) return;
 
 		const updateStickyOffset = () => {
-			const viewportHeight = window.innerHeight;
+			const viewportHeight = window.visualViewport?.height ?? window.innerHeight;
 			const widgetHeight = widget.offsetHeight;
 			const offset = 32;
 			// Negative top is intentional for tall cards: it lets the card scroll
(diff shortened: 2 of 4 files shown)


chrisgen19:
## Code review: PR #146 — dashboard recognition feed / native scroll

Review of branch `issue-145-dashboard-native-scroll` vs `main` (commit `c144b49`). Focus: correctness, regressions, edge cases, tests, performance, docs.

---

### Scope

| File | Change |
|------|--------|
| `app/(dashboard)/dashboard/_components/recognition-feed-widget.tsx` | Remove inner `overflow-y-auto` + `max-h-[calc(100vh-12rem)]` on feed tab panel |
| `app/(dashboard)/dashboard/_components/stats-widget.tsx` | Remove `h-full` from stats surfaces; add `StickyStatsWidget` with measured `top` |
| `app/(dashboard)/dashboard/page.tsx` | Use `StickyStatsWidget` instead of `StatsWidget` |
| `app/(dashboard)/dashboard/loading.tsx` | Align skeleton with sticky stats + order wrappers |

**Intent (as stated):** native page scroll for the recognition feed; keep Recognition Stats visible with a measured sticky offset; align loading skeleton.

---

### 1. Medium — `stickyTop` can go negative (correctness)

**File:** `app/(dashboard)/dashboard/_components/stats-widget.tsx` — `StickyStatsWidget` → `updateStickyOffset`

When `widgetHeight + 2 * offset > viewportHeight`, the code sets `nextTop = viewportHeight - widgetHeight - offset`. If the stats block is **taller than `viewportHeight - offset`**, that value is **negative**. Applying `style={{ top: stickyTop }}` with a negative `top` on `position: sticky` is unreliable and can produce odd sticking or overlap.

**Why more likely after this PR:** Removing `h-full` from the stats card (see §2) lets the card grow with the full “Most Recognized” list, so `offsetHeight` can exceed `window.innerHeight` more often.

**Smallest safe fix:** In the tall-widget branch, clamp, e.g. `Math.max(offset, viewportHeight - widgetHeight - offset)`, so `top` never falls below the intended minimum (likely `32`).

**Future note:** `window.innerHeight` is the viewport; if a fixed desktop chrome or a scrollable `main` region is introduced later, consider measuring the actual scrollport or `visualViewport`.

---

### 2. Medium — Stats height model and inner scroll behavior (behavioral change)

**File:** `app/(dashboard)/dashboard/_components/stats-widget.tsx`

Removing `h-full` from the loaded card, error state, and skeleton means the stats root is **content-sized** in the block axis. The “Most Recognized” `<ol>` still has `flex-1 min-h-0 overflow-y-auto`, but without a definite height from a stretched parent, **`flex-1` often does not bound height**; the list tends to **grow with every row** and **`overflow-y-auto` may never activate**. Scrolling moves to the **document**—aligned with “native scroll” but a **real change** beyond the feed (long leaderboards used to scroll inside the tall card).

**Options:** (a) If full page scroll is intended, simplify the `<ol>` / parent flex classes so they do not imply an inner scroller that rarely exists. (b) If inner scroll is still desired, add an explicit `max-h-*` (or measured cap) instead of relying on `h-full` + grid stretch.

---

### 3. Low — Loading skeleton vs live sticky `top`

**Files:** `loading.tsx` (`lg:top-8`) vs `StickyStatsWidget` (dynamic `stickyTop`).

Initial `32px` matches `top-8`. After resize / `ResizeObserver`, live `stickyTop` can differ in the “tall widget” band. Minor loading vs loaded inconsistency; acceptable or document.

---

### 4. Recognition feed — aligned with goals

**File:** `recognition-feed-widget.tsx`

Removing the magic `100vh - 12rem` cap avoids mismatches with real chrome/safe areas. Parent `app/(dashboard)/layout.tsx` does not trap scroll in `overflow-y-auto` on `main`, so document scrolling is coherent.

**Watch:** If dashboard content is later wrapped in `h-screen overflow-y-auto`, scroll behavior would follow that scrollport.

---

### 5. Dashboard page wiring

**File:** `page.tsx` — `StickyStatsWidget` swap is fine; grid and `lg:` sticky only on large screens remain sensible.

---

### 6. `StickyStatsWidget` — implementation

**Strengths:** `ResizeObserver` guarded; cleanup disconnects observer and removes `resize` listener; observing the wrapper ref picks up post-fetch layout changes.

**Performance:** No scroll listener; low concern.

---

### 7. Security

None; layout-only client changes.

---

### 8. Tests

No coverage for `StickyStatsWidget` / sticky math / unmount cleanup. Optional: unit test with mocked `ResizeObserver` and stubbed dimensions; or Playwright smoke on `/dashboard` at `lg`.

---

### 9. Documentation

PR summary could explicitly note that the **stats leaderboard** may no longer scroll inside the card. A one-line comment above `nextTop` would help maintainers.

---

### Summary

| Severity | Topic | Action |
|----------|--------|--------|
| Medium | Negative `top` when stats very tall | Clamp `stickyTop` |
| Medium | Inner stats list scroll vs `flex-1` / `h-full` removal | Align markup with intent or add `max-h` |
| Low | Skeleton vs dynamic `top` | Optional |
| Low | Tests / docs | Optional |

Overall direction matches the issue; address §1 before edge-case layouts; clarify or adjust §2 if product still wants an internal leaderboard scroll.

chrisgen19:
## Follow-up review — remaining findings (actionable, by severity)

Re-reviewed after the follow-up commit (intentional negative `top` + removal of misleading inner-scroll classes). Earlier “clamp `stickyTop`” / “stale flex-1 + overflow-y” items are **withdrawn / addressed**.

---

### 1. Low — `window.innerHeight` vs mobile dynamic viewport

**File:** `app/(dashboard)/dashboard/_components/stats-widget.tsx` — `updateStickyOffset` in `StickyStatsWidget`

`ResizeObserver` + `resize` do not fire for every **visual viewport** change (e.g. mobile browser chrome showing/hiding). `innerHeight` can jump; `stickyTop` may update a frame late. Polish edge case, not a desktop `lg:sticky` blocker.

**Smallest fix (only if jank shows on real devices):** subscribe to `visualViewport` `resize` / `scroll` where available and base height on `visualViewport.height` with fallback to `innerHeight`.

---

### 2. Low — Empty-state dashed panel lost vertical “fill”

**File:** same — `showEmpty` branch

`flex flex-1` was removed from the inner dashed container (now `flex items-center justify-center` only). With `h-full` already gone, the empty state may render **shorter** than before inside a tall grid row. Visual only.

**Smallest fix:** only if QA/design wants it — e.g. `min-h-48` (or similar) on that dashed `div`, without bringing back inner scroll.

---

**Verdict:** Good to merge from a correctness/regression angle unless you want to chase `visualViewport` in this PR.

chrisgen19:
Addressed the remaining low-severity follow-up items in `b939ee8`:

- Updated `StickyStatsWidget` to use `window.visualViewport?.height` when available and listen to `visualViewport` `resize` / `scroll` events, with `window.innerHeight` as fallback. This keeps the measured sticky offset in sync with dynamic viewport changes.
- Added `min-h-48` to the empty “Most Recognized” dashed panel so the empty state keeps visual weight without restoring any inner scroll behavior.

Verification:
- `bun run lint`
- `git diff --check`

chrisgen19:
## Code Review — PR #146 (recognition feed native scroll + sticky stats)

Re-reviewed against `b939ee8` (current HEAD). The two in-PR self-reviews already captured most concerns and the author resolved them (clamp-intent comment added, misleading inner-scroll classes removed, `visualViewport` support, `min-h-48` empty state). Below are the remaining actionable findings and the context behind each.

---

### 1. Low-Medium — `visualViewport.scroll` listener runs layout reads every scroll frame

**File:** `app/(dashboard)/dashboard/_components/stats-widget.tsx` — `StickyStatsWidget` / `updateStickyOffset`

**What happens:** `visualViewport.addEventListener(\"scroll\", updateStickyOffset)` fires on every page scroll on mobile — iOS Safari dispatches `visualViewport` `scroll` events alongside document scroll, not just on chrome show/hide. Each fire calls `widget.offsetHeight` (a forced style+layout flush) and `setStickyTop(...)`.

**Why it matters:**
- In practice, `visualViewport.height` changes only when the URL bar hides/shows — the `resize` event already covers that case.
- The `scroll` listener therefore adds per-frame layout work on mobile that never produces a different value most of the time.
- `offsetHeight` reads during scroll are a classic source of jank on long pages.
- React will bail out of re-rendering when the numeric state is equal (`Object.is`), but the layout read itself still happens; and on iOS the visual viewport height can jitter by 1px during rubber-band scroll, which *would* trigger real re-renders per frame.

**Smallest safe fix:** Drop the `visualViewport` `scroll` listener; keep `visualViewport.resize` (URL-bar case) and `window.resize` (fallback).

```ts
// remove:
viewport?.addEventListener(\"scroll\", updateStickyOffset);
// and the matching removeEventListener in cleanup
```

If a future bug surfaces that genuinely needs a scroll hook, wrap the work in `requestAnimationFrame` with a leading-edge guard.

---

### 2. Low — Observers run on mobile where `top` is never applied

**File:** same (`StickyStatsWidget`)

**What happens:** `lg:sticky` means `position: sticky` only kicks in at `lg+`; on smaller screens the element is `position: static` and `style.top` has no visual effect. The measurement effect still runs (`ResizeObserver`, window + visualViewport listeners, state updates) on every mount at every breakpoint.

**Why it matters:** Wasted work and unnecessary layout reads on phones. Not a correctness issue, but it compounds with §1 since most real users are on viewports where the effect cannot influence layout.

**Smallest safe fix:** Gate the effect behind a `matchMedia(\"(min-width: 1024px)\")` check — skip installing listeners when the media query is unmatched, and optionally re-run on `change` to cover a user resizing across the breakpoint.

```ts
useEffect(() => {
  const mql = window.matchMedia(\"(min-width: 1024px)\");
  if (!mql.matches) return;
  // …existing setup and cleanup
}, []);
```

For responsive correctness across resize-through-breakpoint, wire `mql.addEventListener(\"change\", …)` and teardown/reinstall on transitions.

---

### 3. Low — No regression test around the sticky math

**File:** no test file for `StickyStatsWidget`

**What happens:** The sticky offset logic is easy to get wrong: the three regimes (short widget → `top = 32`; widget shorter than viewport-2·offset → `top = 32`; widget taller → `top = viewport - height - 32` which is intentionally negative) live entirely in one effect and aren't pinned anywhere.

**Why it matters:** Future refactors (breakpoint gating per §2, rAF throttling per §1, switching away from `visualViewport`) are likely — tests lock the intent behind the code comment so subsequent changes don't silently invert the behavior.

**Smallest safe fix (optional follow-up, not blocking):** A small Vitest with a stubbed `ResizeObserver` and controllable `window.innerHeight` / `window.visualViewport.height` to cover:
- short widget → renders with `style.top: 32px`
- tall widget → renders with a negative `top` value matching `viewport - height - offset`
- changing `visualViewport.height` via `resize` triggers a recomputation

---

### Informational (no action required)

- **Skeleton vs live `top` drift.** `loading.tsx` uses `lg:top-8` while `StickyStatsWidget` may compute a dynamic (possibly negative) `top` once mounted. Only visible when the stats card exceeds viewport height, and the flash is imperceptible in practice — fine to leave.
- **Negative `stickyTop` is intentional.** The code comment and the author's follow-up already explain this. Works with `position: sticky` because the element still pins once its `top` edge intersects the viewport; a negative `top` pulls the pin further up so the card's bottom stays reachable on tall layouts.
- **Recognition feed `max-h-[calc(100vh-12rem)]` removal** is the right call — that cap was fighting real chrome/safe-area heights. Parent `main` does not trap scroll (`overflow-y-auto` is not set on `main`), so document scroll works end-to-end, consistent with the PR's stated intent.
- **`h-full` removal on stats surfaces** is coherent with the native-scroll model: the card is now content-sized, the leaderboard grows with rows, and the page owns scrolling. The `min-h-48` on the empty state keeps visual weight without reintroducing inner scroll.

---

### Security

No new concerns. All changes are client-side layout/CSS; no user-controlled HTML, no new data flow, auth/session paths untouched.

---

### Performance recap

- \"Native scroll\" direction is a net performance win — the browser owns scroll, no custom scroll listeners on the feed.
- `StickyStatsWidget`'s only residual performance cost is §1 (per-frame layout read on mobile). Removing the `scroll` listener brings this to essentially zero cost.
- `ResizeObserver` + `resize` is the right primitive for height-against-viewport computation; no re-architecting needed.

---

### Summary

| # | Severity | Topic | Action |
|---|---|---|---|
| 1 | Low-Medium | `visualViewport.scroll` listener is redundant and triggers layout reads per scroll frame | Drop the `scroll` listener; keep `resize` |
| 2 | Low | Effect runs on mobile where `lg:sticky` doesn't apply | Gate setup on `matchMedia(\"(min-width: 1024px)\")` |
| 3 | Low | Missing tests for `StickyStatsWidget` sticky math | Optional follow-up; not blocking |

**Verdict:** The PR achieves its stated intent and the earlier self-reviews already handled the substantive correctness issues (clamp intent, stale inner-scroll classes, visual viewport, empty-state weight). Safe to merge — §1 and §2 are polish that can ship in this PR or a quick follow-up; §3 is nice-to-have.

chrisgen19:
Addressed the remaining valid performance findings from https://github.com/chrisgen19/access-group-staff/pull/146#issuecomment-4309331774 in `4f260f2`:

- Removed the `visualViewport` `scroll` listener, so sticky offset recalculation no longer performs layout reads on every mobile scroll frame. `visualViewport.resize` and `window.resize` remain for viewport height changes.
- Gated `StickyStatsWidget` measurement behind `matchMedia("(min-width: 1024px)")`, matching the `lg:sticky` breakpoint. The effect now installs listeners only on desktop and tears them down / reinstalls when crossing the breakpoint.
- Left the sticky math test as optional follow-up; no behavior gap required it for this PR.

Verification:
- `bun run lint`
- `bunx tsc --noEmit`
- `git diff --check`

PR #148, opened 2026-04-24
Optimise recognition feed: limit + cursor pagination (closes #147)

### LINKED ISSUE #147: Optimise recognition feed: reduce initial payload + cursor-based infinite scroll
## Problem

`RecognitionFeed` (`app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx`) currently fetches **up to 50 cards** on every mount via `GET /api/recognition`, regardless of viewport or surface. The same component is used by:

- The **dashboard widget** (`recognition-feed-widget.tsx`) — a small card where users typically only glance at the top few.
- The **full recognition page** (`/dashboard/recognition`) — where users may want to browse history.

Additionally, the API eagerly fetches **every reaction row** for all 50 cards with the reactor user joined in (`app/api/recognition/route.ts:186-202`). For cards nobody interacts with, this is wasted work.

Most users glance at the dashboard widget and leave — they pay the full 50-card cost for a handful of visible cards.

### 1. Reduce the initial payload on the dashboard widget
Add a `limit` query param to `/api/recognition` (default 50 to stay backward-compatible). The dashboard widget passes `limit=10`. No UX change — it already only shows a vertical stack and users scroll the page, not the widget.

### 2. Cursor-based pagination + infinite scroll on the full recognition page
Replace the current \`useQuery\` + \`take: 50\` pattern on \`/dashboard/recognition\` with:

- **API:** cursor pagination (\`?cursor=<id>&limit=10\`) returning \`{ data, nextCursor }\`. Cursor > \`skip/offset\` because \`skip\` forces Postgres to scan past rows on every page.
- **Client:** \`useInfiniteQuery\` from TanStack Query, with an IntersectionObserver sentinel that calls \`fetchNextPage()\` when it enters the viewport (with \`rootMargin\` prefetch).
- **Fallback:** a visible \"Load more\" button for keyboard / screen-reader users — don't rely on scroll alone.

## Non-goals (explicitly out of scope)
- Server-side \`use cache\` + \`cacheTag\` — separate issue if needed.
- Trimming the reaction join (lazy-loading reactors on popover open) — separate issue.
- Changes to the admin paginated branch (\`?paginated=true\`) — already paginated correctly.

## Regression guardrails
- \`/api/recognition\` remains backward-compatible: no \`limit\`/\`cursor\` params = current behaviour (\`take: 50\`, no cursor).
- All existing callers (\`recognition-feed.tsx\`, admin page, any other consumers) must be audited before changing defaults.
- Reaction summary shape, card shape, and response envelope (\`{ success, data }\`) unchanged.
- Admin paginated branch (\`paginated=true\`) untouched.
- Add E2E coverage for: widget loads ≤10 cards, full page loads first 10 + loads more on scroll, \"Load more\" button works.
- Add unit tests for cursor logic in the API route (first page, middle page, last page, empty result).

## Acceptance criteria
- [ ] Dashboard widget fetches 10 cards on initial load (verified in Network tab).
- [ ] \`/dashboard/recognition\` fetches 10 cards on initial load and appends 10 more on scroll.
- [ ] \"Load more\" button visible and functional for non-scroll users.
- [ ] No regression in create / edit / react / comment flows (all existing E2E passes).
- [ ] No change in card ordering (newest first by \`createdAt\`).
- [ ] Admin paginated view unaffected.

chrisgen19:
## Summary
- `GET /api/recognition` gains backward-compatible `limit` (1–50, default 50) and `cursor` query params; adds `nextCursor` to the response. Admin `paginated=true` and `export=true` branches untouched.
- Dashboard widget now fetches 10 cards instead of 50 (5× smaller initial payload for the most-visited surface).
- Inbox `received` and `sent` pages switch to `useInfiniteQuery` with an IntersectionObserver sentinel (400px prefetch) + accessible "Load more" fallback button.
- New `useIntersectionObserver` hook (no new dependency).
- Query key stays prefix-compatible (`["recognition-cards", ...]`), so existing `invalidateQueries` in `recognition-form` and `recognition-table` keep working against both static and infinite queries.

## Regression guardrails
- `/api/recognition` with no new params = previous behaviour (50 cards, no cursor) plus an additive `nextCursor` field old clients ignore.
- `orderBy` now `[{ createdAt: desc }, { id: desc }]` to guarantee stable cursors across same-ms rows.
- Reaction summary join uses the trimmed page (not the `+1` overfetch row).
- Admin `paginated=true` branch shape (page/pageSize/total/totalPages) unchanged.
- React-query invalidations from card create (`recognition-form`) and admin delete (`recognition-table`) already use the `["recognition-cards"]` prefix, which matches both `useQuery` and `useInfiniteQuery` — verified by inspection.

## Tests added
- `app/api/recognition/route.test.ts` — 8 cases: defaults, limit+1 overfetch, cursor forwarding, clamping, empty result, reaction batch scope.
- `hooks/use-intersection-observer.test.tsx` — 4 cases: observe, callback, disabled, disconnect on unmount.
- `app/(dashboard)/dashboard/recognition/_components/recognition-feed.test.tsx` — 3 cases: first page, Load more appends page 2 with cursor, no button when `nextCursor` null.

## Test plan
- [ ] Dashboard: widget loads 10 cards, tab `Public` ⇄ `My Department` still refetches.
- [ ] Inbox received: first 10 cards render, scroll to bottom → next 10 appear; "Load more" button works when scroll is impractical.
- [ ] Inbox sent: same as received.
- [ ] Create a new card → appears at the top of infinite feed (invalidation through query prefix).
- [ ] React/comment on a card → still works (per-card invalidation path unchanged).
- [ ] Admin `/dashboard/recognition/all` paginated table unaffected.
- [ ] CSV export (admin) unaffected.
- [ ] Unread "New" badge still flags unread cards across page boundaries.

Closes #147

### DIFF
--- a/app/(dashboard)/dashboard/_components/recognition-feed-widget.tsx
+++ b/app/(dashboard)/dashboard/_components/recognition-feed-widget.tsx
@@ -71,6 +71,7 @@ export function RecognitionFeedWidget({ currentUserId, isAdmin }: RecognitionFee
 					emptyTitle={currentTab.emptyTitle}
 					emptyDescription={currentTab.emptyDescription}
 					limit={10}
+					infinite
 				/>
 			</div>
 		</div>
--- a/app/(dashboard)/dashboard/_components/recognition-feed-widget.tsx
+++ b/app/(dashboard)/dashboard/_components/recognition-feed-widget.tsx
@@ -70,6 +70,7 @@ export function RecognitionFeedWidget({ currentUserId, isAdmin }: RecognitionFee
 					isAdmin={isAdmin}
 					emptyTitle={currentTab.emptyTitle}
 					emptyDescription={currentTab.emptyDescription}
+					limit={10}
 				/>
 			</div>
 		</div>
--- a/app/(dashboard)/dashboard/recognition/_components/recognition-feed-client.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/recognition-feed-client.tsx
@@ -33,6 +33,8 @@ export function RecognitionFeedClient({
 				emptyTitle={emptyTitle}
 				emptyDescription={emptyDescription}
 				onShare={setShareCardId}
+				limit={10}
+				infinite
 			/>
 			<ShareDialog
 				open={!!shareCardId}
--- a/app/(dashboard)/dashboard/recognition/_components/recognition-feed.test.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/recognition-feed.test.tsx
@@ -21,8 +21,9 @@ vi.mock("./recognition-card-mini", () => ({
 	RecognitionCardMini: () => <div />,
 }));
 
+const intersectingState = { current: false };
 vi.mock("@/hooks/use-intersection-observer", () => ({
-	useIntersectionObserver: () => ({ ref: () => {}, isIntersecting: false }),
+	useIntersectionObserver: () => ({ ref: () => {}, isIntersecting: intersectingState.current }),
 }));
 
 import { RecognitionFeed } from "./recognition-feed";
@@ -50,6 +51,7 @@ const fetchMock = vi.fn();
 
 beforeEach(() => {
 	fetchMock.mockReset();
+	intersectingState.current = false;
 	vi.stubGlobal("fetch", fetchMock);
 });
 
@@ -154,6 +156,46 @@ describe("RecognitionFeed (infinite)", () => {
 					?.status,
 			).toBe("error"),
 		);
+	});
+
+	test("auto-fetches next page when sentinel intersects after a user scroll", async () => {
+		fetchMock
+			.mockResolvedValueOnce({
+				ok: true,
+				json: async () => ({
+					success: true,
+					data: [makeCard("a"), makeCard("b")],
+					nextCursor: "b",
+				}),
+			})
+			.mockResolvedValueOnce({
+				ok: true,
+				json: async () => ({
+					success: true,
+					data: [makeCard("c")],
+					nextCursor: null,
+				}),
+			});
+
+		const client = new QueryClient({ defaultOptions: { queries: { retry: false } } });
+		render(<RecognitionFeed filter="received" limit={2} infinite />, { wrapper: wrapper(client) });
+		await waitFor(() => expect(screen.getByText("message a")).toBeInTheDocument());
+
+		// Sentinel is in view but user hasn't scrolled yet — no second fetch.
+		await act(async () => {
+			intersectingState.current = true;
+			window.dispatchEvent(new Event("resize"));
+		});
+		expect(fetchMock).toHaveBeenCalledTimes(1);
+
+		// First user scroll flips the gate; sentinel still intersecting → fetch fires.
+		await act(async () => {
+			window.dispatchEvent(new Event("scroll"));
+		});
+		await waitFor(() => expect(fetchMock).toHaveBeenCalledTimes(2));
+
+		const secondCall = fetchMock.mock.calls[1]?.[0] as string;
+		expect(secondCall).toContain("cursor=b");
 	});
 
 	test("hides Load more when nextCursor is null on first page", async () => {
--- a/app/(dashboard)/dashboard/recognition/_components/recognition-feed.test.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/recognition-feed.test.tsx
@@ -1,0 +1,133 @@
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { act, cleanup, render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import type { ReactNode } from "react";
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+
+vi.mock("@/hooks/use-unread-card-ids", () => ({
+	useUnreadCardIds: () => ({ unreadCardIds: new Set<string>() }),
+}));
+
+vi.mock("@/stores/use-preferences-store", () => ({
+	usePreferencesStore: (selector: (s: { cardView: string; cardSize: string }) => unknown) =>
+		selector({ cardView: "list", cardSize: "md" }),
+}));
+
+vi.mock("./card-interaction-bar", () => ({
+	CardInteractionBar: () => <div data-testid="interaction-bar" />,
+}));
+
+vi.mock("./recognition-card-mini", () => ({
+	RecognitionCardMini: () => <div />,
+}));
+
+vi.mock("@/hooks/use-intersection-observer", () => ({
+	useIntersectionObserver: () => ({ ref: () => {}, isIntersecting: false }),
+}));
+
+import { RecognitionFeed } from "./recognition-feed";
+
+function makeCard(id: string) {
+	return {
+		id,
+		message: `message ${id}`,
+		date: "2026-04-22T00:00:00.000Z",
+		createdAt: "2026-04-22T00:00:00.000Z",
+		sender: { id: "s", firstName: "Sa", lastName: "Se", avatar: null, position: null },
+		recipient: { id: "r", firstName: "Re", lastName: "Ri", avatar: null, position: null },
+		interactionCounts: { reactions: 0, comments: 0 },
+		reactionSummary: [],
+	};
+}
+
+function wrapper(client: QueryClient) {
+	return ({ children }: { children: ReactNode }) => (
+		<QueryClientProvider client={client}>{children}</QueryClientProvider>
+	);
+}
+
+const fetchMock = vi.fn();
+
+beforeEach(() => {
+	fetchMock.mockReset();
+	vi.stubGlobal("fetch", fetchMock);
+});
+
+afterEach(() => {
+	cleanup();
+	vi.unstubAllGlobals();
+});
+
+describe("RecognitionFeed (infinite)", () => {
+	test("loads first page and renders Load more when nextCursor present", async () => {
+		fetchMock.mockResolvedValueOnce({
+			ok: true,
+			json: async () => ({
+				success: true,
+				data: [makeCard("a"), makeCard("b")],
+				nextCursor: "b",
+			}),
+		});
+
+		const client = new QueryClient({ defaultOptions: { queries: { retry: false } } });
+		render(<RecognitionFeed filter="received" limit={2} infinite />, { wrapper: wrapper(client) });
+
+		await waitFor(() => expect(screen.getByText("message a")).toBeInTheDocument());
+		expect(screen.getByRole("button", { name: /load more/i })).toBeInTheDocument();
+
+		const call = fetchMock.mock.calls[0]?.[0] as string;
+		expect(call).toContain("filter=received");
+		expect(call).toContain("limit=2");
+		expect(call).not.toContain("cursor=");
+	});
+
+	test("clicking Load more appends next page with cursor", async () => {
+		fetchMock
+			.mockResolvedValueOnce({
+				ok: true,
+				json: async () => ({
+					success: true,
+					data: [makeCard("a"), makeCard("b")],
+					nextCursor: "b",
+				}),
+			})
+			.mockResolvedValueOnce({
+				ok: true,
+				json: async () => ({
+					success: true,
+					data: [makeCard("c")],
+					nextCursor: null,
+				}),
+			});
+
+		const client = new QueryClient({ defaultOptions: { queries: { retry: false } } });
+		render(<RecognitionFeed filter="received" limit={2} infinite />, { wrapper: wrapper(client) });
+
+		await waitFor(() => expect(screen.getByText("message a")).toBeInTheDocument());
+		await act(async () => {
+			await userEvent.click(screen.getByRole("button", { name: /load more/i }));
+		});
+		await waitFor(() => expect(screen.getByText("message c")).toBeInTheDocument());
+
+		expect(screen.queryByRole("button", { name: /load more/i })).not.toBeInTheDocument();
+		const secondCall = fetchMock.mock.calls[1]?.[0] as string;
+		expect(secondCall).toContain("cursor=b");
+	});
+
+	test("hides Load more when nextCursor is null on first page", async () => {
+		fetchMock.mockResolvedValueOnce({
+			ok: true,
+			json: async () => ({
+				success: true,
+				data: [makeCard("a")],
+				nextCursor: null,
+			}),
+		});
+
+		const client = new QueryClient({ defaultOptions: { queries: { retry: false } } });
+		render(<RecognitionFeed filter="sent" limit={10} infinite />, { wrapper: wrapper(client) });
+
+		await waitFor(() => expect(screen.getByText("message a")).toBeInTheDocument());
+		expect(screen.queryByRole("button", { name: /load more/i })).not.toBeInTheDocument();
+	});
+});
--- a/app/(dashboard)/dashboard/recognition/_components/recognition-feed.test.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/recognition-feed.test.tsx
@@ -58,6 +58,30 @@ afterEach(() => {
 	vi.unstubAllGlobals();
 });
 
+describe("RecognitionFeed (static)", () => {
+	test("fetches once, forwards limit, and renders no Load more", async () => {
+		fetchMock.mockResolvedValueOnce({
+			ok: true,
+			json: async () => ({
+				success: true,
+				data: [makeCard("a")],
+				nextCursor: null,
+			}),
+		});
+
+		const client = new QueryClient({ defaultOptions: { queries: { retry: false } } });
+		render(<RecognitionFeed filter="all" limit={25} />, { wrapper: wrapper(client) });
+
+		await waitFor(() => expect(screen.getByText("message a")).toBeInTheDocument());
+		expect(screen.queryByRole("button", { name: /load more/i })).not.toBeInTheDocument();
+
+		const call = fetchMock.mock.calls[0]?.[0] as string;
+		expect(call).toContain("limit=25");
+		expect(call).not.toContain("filter=all");
+		expect(fetchMock).toHaveBeenCalledTimes(1);
+	});
+});
+
 describe("RecognitionFeed (infinite)", () => {
 	test("loads first page and renders Load more when nextCursor present", async () => {
 		fetchMock.mockResolvedValueOnce({
@@ -112,6 +136,24 @@ describe("RecognitionFeed (infinite)", () => {
 		expect(screen.queryByRole("button", { name: /load more/i })).not.toBeInTheDocument();
 		const secondCall = fetchMock.mock.calls[1]?.[0] as string;
 		expect(secondCall).toContain("cursor=b");
+	});
+
+	test("throws when body.success is false (consistent error envelope)", async () => {
+		fetchMock.mockResolvedValueOnce({
+			ok: true,
+			json: async () => ({ success: false, error: "nope" }),
+		});
+
+		const client = new QueryClient({ defaultOptions: { queries: { retry: false } } });
+		render(<RecognitionFeed filter="received" limit={5} infinite />, { wrapper: wrapper(client) });
+
+		// Query errors cause isPending to finish without populating data; skeletons linger on error.
+		await waitFor(() =>
+			expect(
+				client.getQueryState(["recognition-cards", "received", { limit: 5, infinite: true }])
+					?.status,
+			).toBe("error"),
+		);
 	});
 
 	test("hides Load more when nextCursor is null on first page", async () => {
--- a/app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useInfiniteQuery, useQuery } from "@tanstack/react-query";
+import { type InfiniteData, useInfiniteQuery, useQuery } from "@tanstack/react-query";
 import { ArrowRight, Eye, Heart, Pencil, Share2 } from "lucide-react";
 import { useRouter } from "next/navigation";
 import { useEffect } from "react";
@@ -21,6 +21,14 @@ interface FeedPage {
 	success: boolean;
 	data: RecognitionCard[];
 	nextCursor: string | null;
+}
+
+async function fetchFeedPage(url: string): Promise<FeedPage> {
+	const res = await fetch(url);
+	if (!res.ok) throw new Error("Failed to fetch recognition cards");
+	const body = (await res.json()) as FeedPage;
+	if (!body.success) throw new Error("Failed to fetch recognition cards");
+	return body;
 }
 
 function buildFeedUrl(filter: FeedFilter, limit: number, cursor?: string | null) {
@@ -152,11 +160,7 @@ function RecognitionFeedStatic({
 }: RecognitionFeedProps) {
 	const { data, isPending } = useQuery<FeedPage>({
 		queryKey: ["recognition-cards", filter, { limit, infinite: false }],
-		queryFn: async () => {
-			const res = await fetch(buildFeedUrl(filter, limit));
-			if (!res.ok) throw new Error("Failed to fetch recognition cards");
-			return res.json();
-		},
+		queryFn: () => fetchFeedPage(buildFeedUrl(filter, limit)),
 		staleTime: 30_000,
 	});
 
@@ -194,17 +198,13 @@ function RecognitionFeedInfinite({
 	const { data, isPending, hasNextPage, isFetchingNextPage, fetchNextPage } = useInfiniteQuery<
 		FeedPage,
 		Error,
-		{ pages: FeedPage[] },
+		InfiniteData<FeedPage>,
 		readonly unknown[],
 		string | null
 	>({
 		queryKey: ["recognition-cards", filter, { limit, infinite: true }],
 		initialPageParam: null,
-		queryFn: async ({ pageParam }) => {
-			const res = await fetch(buildFeedUrl(filter, limit, pageParam));
-			if (!res.ok) throw new Error("Failed to fetch recognition cards");
-			return res.json();
-		},
+		queryFn: ({ pageParam }) => fetchFeedPage(buildFeedUrl(filter, limit, pageParam)),
 		getNextPageParam: (last) => last.nextCursor ?? undefined,
 		staleTime: 30_000,
 	});
--- a/app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
+++ b/app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx
@@ -1,16 +1,36 @@
 "use client";
 
-import { useQuery } from "@tanstack/react-query";
+import { useInfiniteQuery, useQuery } from "@tanstack/react-query";
 import { ArrowRight, Eye, Heart, Pencil, Share2 } from "lucide-react";
 import { useRouter } from "next/navigation";
+import { useEffect } from "react";
 import { SkeletonCard, SkeletonLine } from "@/components/shared/skeleton-primitives";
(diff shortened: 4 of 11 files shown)


chrisgen19:
## Code review (correctness, regressions, security, tests)

Focused on [#147](https://github.com/chrisgen19/access-group-staff/issues/147) goals: smaller initial payloads + cursor-based infinite scroll.

### Findings (by severity)

#### 1. Invalid or non-matching `cursor` can surface as 500 instead of a safe client error

**Files:** `app/api/recognition/route.ts`  
**Why it matters:** `cursor` is user-controlled query input. If Prisma cannot resolve the cursor row (unknown id, wrong type for the column, or a row that does not satisfy `where` in some Prisma versions), `findMany` can throw. That is swallowed by the broad `catch` and returned as **500**, which is noisy for monitoring, easier to abuse for log noise, and worse UX than a controlled 400/empty page.  
**Smallest safe fix:** Narrow the `catch` (or wrap the `findMany` in a small try/catch) and map Prisma “record not found” / invalid cursor cases to **400** with `{ success: false, data: [], nextCursor: null }` (or ignore cursor and treat as first page). Optionally validate `cursor` against the same UUID/string format you use for `RecognitionCard.id` in `schema.prisma` before calling Prisma.

---

#### 2. Fractional `limit` is not normalized to an integer

**Files:** `app/api/recognition/route.ts` (limit parsing ~`174–178`, `take: limit + 1`)  
**Why it matters:** A request like `?limit=10.7` keeps `limit` at `10.7`, so `take` becomes `11.7`. Database drivers often coerce, but this is undefined behavior and could break or warn depending on driver/Prisma.  
**Smallest safe fix:** After clamping, use an integer, e.g. `const limit = Math.floor(Math.min(50, Math.max(1, ...)))` (or `parseInt` with explicit radix after validating finite).

---

#### 3. Feed `queryFn` does not verify `success` in the JSON body

**Files:** `app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx` (`RecognitionFeedStatic` and `RecognitionFeedInfinite` `queryFn`)  
**Why it matters:** You only check `res.ok`. If the route ever returns **200** with `{ success: false, ... }`, React Query will treat it as success and may render empty or inconsistent UI without a proper error state. Low probability today because this handler usually uses non-200 for auth errors.  
**Smallest safe fix:** After `res.json()`, if `!body.success`, `throw new Error(...)` (or parse a typed error) so the query goes to error/retry behavior.

---

#### 4. Tests miss several high-value edge cases for the new API contract

**Files:** `app/api/recognition/route.test.ts`  
**Why it matters:** Regression protection for the new surface (`limit` + `cursor`) is good for the happy path, but gaps leave the two issues above unguarded.  
**Smallest safe fix:** Add cases for: (a) **non-integer limit** → expect integer `take`; (b) **invalid cursor** → assert status/body (once you define desired behavior in fix #1); optionally (c) `limit` as non-numeric string (already effectively defaults—assert `take`).

---

#### 5. `useInfiniteQuery` generics are imprecise (maintainability only)

**Files:** `app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx` (~`194–200`)  
**Why it matters:** The third generic is written as `{ pages: FeedPage[] }` instead of `InfiniteData<FeedPage>`. Runtime is fine; TypeScript may lie about `pageParams` and related types.  
**Smallest safe fix:** Use `InfiniteData<FeedPage>` from `@tanstack/react-query` as the `TData` generic (or omit extra generics and let inference work).

---

#### 6. No automated coverage for the “static” (non-infinite) branch

**Files:** `app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx` (`RecognitionFeedStatic`), `recognition-feed.test.tsx`  
**Why it matters:** Today every in-repo usage passes `infinite` (`recognition-feed-widget.tsx`, `recognition-feed-client.tsx`), so the static path is easy to break silently if someone later uses `RecognitionFeed` without `infinite`.  
**Smallest safe fix:** One small test: `infinite` omitted/false → single fetch, no “Load more”, `limit` forwarded—**or** document in a code comment that static mode is legacy/public API only (prefer a test if you want a hard guard).

---

### Positive checks (not issues)

- **Stable ordering** `[{ createdAt: "desc" }, { id: "desc" }]` plus `nextCursor` from the last returned row matches common cursor patterns for tie-breaking.
- **Reaction batching** scoped to returned card ids (not the overfetch row) is correct and covered in `route.test.ts`.
- **Query invalidation** via `queryKey: ["recognition-cards"]` still matches both `useQuery` and `useInfiniteQuery` keys in this codebase (`recognition-form.tsx`, `recognition-table.tsx`).
- **TanStack Query v5** infinite refetch replays pages using fresh `getNextPageParam` chains, so “create card then invalidate” is much less likely to leave inconsistent page stacks than older patterns.
- **Admin** `paginated=true` / `export=true` gate remains before the new branch in `route.ts`, so admin shape and CSV paths stay isolated from cursor logic.

---

### Issue #147 alignment

The PR matches the issue intent: **smaller first paint** (widget and inbox use `limit={10}`) and **cursor pagination** on the API with **infinite loading** in the UI, plus tests on the API, hook, and feed. Remaining risk is mostly **input hardening** (`cursor`, fractional `limit`) and **slightly stronger client/server contract checks** (`success` flag, extra tests).

---

*Posted from local review of the branch / workspace.*

chrisgen19:
## Code review (updated — current branch / workspace)

Review focused on correctness, regressions, security, edge cases, tests, breaking changes, performance, and documentation.

---

### 1. Any `findMany` failure is reported as **400 Invalid cursor** (including first page, no cursor)

**Files:** `app/api/recognition/route.ts` (the `findMany(...).catch(() => null)` block and the `if (rows === null)` branch)

**Why it matters:** The `.catch(() => null)` is attached to **every** non-admin feed `findMany`, not only when `cursor` is set. If Prisma rejects (timeouts, pool exhaustion, transient DB errors, etc.), the handler always returns **400** with `"Invalid cursor"`. That is the wrong status and message for infrastructure failures, and it is wrong even when the client sent **no** cursor (first page). It also makes monitoring/alerts harder because real outages look like client bad input.

**Smallest safe fix:** Only treat `null` as an invalid cursor when `cursor` was actually provided, e.g. wrap the cursor path in its own try/catch, or `catch` only when `cursor` is truthy and rethrow / fall through to the outer `catch` when `!cursor`. Prefer mapping **only** Prisma “record not found / invalid cursor” codes to **400** and letting everything else hit the existing **500** path.

---

### 2. The test “maps Prisma findMany errors … to 400” locks in the overly broad behavior above

**Files:** `app/api/recognition/route.test.ts` (`maps Prisma findMany errors (e.g. bad cursor) to 400`)

**Why it matters:** The test uses `mockRejectedValueOnce(new Error("bad cursor"))` with `?cursor=abc`. That does not distinguish Prisma cursor errors from **any** rejection. After tightening the route (finding #1), this test should assert the intended cases only (e.g. invalid cursor → 400; generic DB error without cursor → 500).

**Smallest safe fix:** Split or adjust tests after the route fix: one case for invalid cursor (400), one for `findMany` rejection with **no** cursor (500), if you keep both paths testable.

---

### 3. Residual product/edge risk: bogus cursor that Prisma accepts but returns an empty page

**Files:** `app/api/recognition/route.ts` (cursor + `where` interaction)

**Why it matters:** Depending on Prisma/version, a syntactically valid id that is not in the filtered result set may yield an **empty** page with `nextCursor: null`, which looks like “end of feed” rather than “bad cursor”. Low frequency in normal clients (they only send ids from `nextCursor`), but possible with tampered URLs.

**Smallest safe fix:** Only if you see this in practice: document as acceptable, or validate that the cursor row exists and matches `where` before paginating (extra query — tradeoff vs. clarity).

---

### What already looks solid (no new issues from this pass)

- Integer `limit` via `Math.floor`, fractional limit test, oversized cursor ignored, reaction query scoped to returned ids, stable `orderBy`, `fetchFeedPage` enforcing `success`, `InfiniteData<FeedPage>`, static + infinite + `success: false` feed tests, admin/export branch still gated earlier in the handler.

---

### Docs gap (low)

**Files:** repo-level API docs (if you maintain any)

**Why:** New `limit`, `cursor`, and `nextCursor` are not self-documenting for other clients.

**Smallest fix:** A short note in whatever you use for HTTP contracts (README fragment, internal wiki, or OpenAPI if you add it later).

---

**Summary:** The main actionable problem in the current code is **#1** (unconditional `.catch` + 400 for all `findMany` failures). Addressing that also drives how you revise **#2**. Earlier review items (fractional `limit`, cursor length, `success` check on feed, `InfiniteData`, static feed tests) appear addressed in-tree.

*Posted from Cursor / local workspace review.*

chrisgen19:
## Review fixes applied

Addressed both review passes ([first](https://github.com/chrisgen19/access-group-staff/pull/148#issuecomment-4309771740), [second](https://github.com/chrisgen19/access-group-staff/pull/148#issuecomment-4309904974)).

### Commit `91a285f` — first review pass

| # | Finding | Fix |
|---|---|---|
| 1 | Invalid cursor could surface as 500 | Cap cursor at 64 chars + guarded try/catch → 400 `{ success: false, error: "Invalid cursor" }` |
| 2 | Fractional `limit` not floored | `Math.floor(...)` — `?limit=10.7` now produces `take: 11` |
| 3 | `queryFn` didn't verify `body.success` | Shared `fetchFeedPage` helper throws when `!body.success`, so 200 + `{ success: false }` surfaces as a React Query error |
| 4 | Test gaps | +3 API cases (fractional limit, oversized cursor, Prisma throws), +1 static-mode regression guard, +1 success-envelope |
| 5 | `{ pages: FeedPage[] }` generic | Replaced with `InfiniteData<FeedPage>` from `@tanstack/react-query` |
| 6 | No static-branch test | Added — single fetch, `limit` forwarded, no "Load more" |

### Commit `c3f78c7` — second review pass

| # | Finding | Fix |
|---|---|---|
| 1 | `.catch(() => null)` mapped **every** findMany failure to 400 — including first-page requests with no cursor, and infra errors (pool exhaustion, timeouts) | Explicit try/catch: 400 only when `cursor` was supplied; otherwise rethrow to outer 500. Restores correct monitoring signal for outages. |
| 2 | Test locked in the overly broad behaviour | Split into two cases: `findMany throws + cursor` → 400; `findMany throws + no cursor` → 500 |

### Not addressed (by reviewer's own recommendation)

- **Tampered but syntactically valid cursor → empty page looks like end-of-feed** — flagged as "only if you see this in practice"; acceptable tradeoff.
- **API docs** — no existing HTTP-contract docs in the repo to update.

### Verification

- `bun run tsc --noEmit` — clean
- `bun run biome check .` — clean
- `bun run vitest run` — 225/225 passing (12 API, 4 hook, 5 feed, rest unrelated)

Ready for another look.

chrisgen19:
## Remaining findings (post-fix review)

Earlier review items (fractional `limit`, oversized cursor, `fetchFeedPage` + `success`, `InfiniteData` typing, static feed tests, and **400 vs 500** when `findMany` fails **without** a cursor) look **addressed** in the current branch — nice work.

What follows is **non-blocking** polish / follow-ups only.

---

### 1. With `cursor` present, any `findMany` rejection still maps to **400 Invalid cursor**

**File:** `app/api/recognition/route.ts` (inner `try/catch` around `recognitionCard.findMany`)

**Why it matters:** The important regression (**no** cursor → mis-400) is fixed by rethrowing when `!cursor`. When **`cursor` is set**, any thrown error still returns **400** with `"Invalid cursor"`. A **DB timeout**, pool exhaustion, or other infra failure during “load next page” is therefore still labeled as a client cursor error, which can confuse operators and dashboards.

**Smallest safe fix (optional):** Map **only** Prisma errors that truly mean an invalid / missing cursor record (e.g. known `code` / `meta` from `PrismaClientKnownRequestError`) to **400**; for all other errors with `cursor` set, **rethrow** so the outer handler returns **500**.

---

### 2. Rare UX edge: tampered cursor may look like “end of feed”

**File:** `app/api/recognition/route.ts` (cursor + `where` semantics)

**Why it matters:** Depending on Prisma/version, a syntactically valid id that does not belong in the filtered feed might yield an **empty** page and **`nextCursor: null`**, which reads like a normal end-of-list. Normal clients only send ids from `nextCursor`, so this is low frequency.

**Smallest safe fix (optional):** Document as acceptable, or add a cheap existence check that the cursor row matches `where` if product wants a distinct error/empty state.

---

### 3. Documentation gap (low)

**Why:** `limit`, `cursor`, and `nextCursor` on `GET /api/recognition` are easy for the next integrator to miss if nothing describes the contract.

**Smallest fix:** A short paragraph in whatever you use for API notes (README, OpenAPI, internal doc).

---

### 4. Local Vitest / CI note (environment)

In one local environment, `vitest run` failed at config load (`ERR_REQUIRE_ESM` / `vitest.config.ts`). That appears **tooling/environment-specific**, not tied to this PR. **GitHub Actions (or your usual CI)** remains the authoritative test gate before merge.

---

*Posted from Cursor — remaining findings only.*

chrisgen19:
## Post-fix review — responses

Addressing [this comment](https://github.com/chrisgen19/access-group-staff/pull/148#issuecomment-4310048703).

### Commit `085ce91` — fix

**#1 Cursor-set infra errors mapped to 400** — ✅ fixed.

Narrowed the 400 mapping to Prisma-typed errors only:

```ts
const isCursorError =
  !!cursor &&
  (err instanceof Prisma.PrismaClientValidationError ||
   err instanceof Prisma.PrismaClientKnownRequestError);
if (isCursorError) {
  return Response.json({ success: false, error: "Invalid cursor" }, { status: 400 });
}
throw err;
```

A DB timeout / pool exhaustion / connection failure during "load next page" now correctly hits the outer 500 handler, restoring monitoring signal for infrastructure incidents.

Tests updated accordingly:
- `Prisma.PrismaClientValidationError` + cursor → **400**
- Generic `Error` (infra) + cursor → **500** (new)
- `Prisma.PrismaClientValidationError` + no cursor → **500**

### Not addressed (reviewer flagged as optional / out of scope)

- **#2 Tampered cursor → empty page** — reviewer: "document as acceptable". Acceptable tradeoff; normal clients only pass ids from `nextCursor`.
- **#3 API docs gap** — no existing HTTP-contract docs in the repo to update.
- **#4 Vitest config / ERR_REQUIRE_ESM** — reviewer: "tooling/environment-specific, not tied to this PR". Vitest runs clean in my env and in CI.

### Verification

- `bun run tsc --noEmit` — clean
- `bun run biome check .` — clean
- `bun run vitest run` — **226/226** passing (13 API cases now)

chrisgen19:
## Final review status\n\nRe-reviewed the latest commits on this PR after the follow-up fixes.\n\n- All previously raised actionable findings are addressed.\n- CI checks are green (lint/unit, migrations replay, e2e, security, preview).\n- No new correctness, regression, or security blockers found in the updated code paths.\n\nRemaining items are non-blocking polish only (API contract docs / optional cursor edge-case clarification).\n\nFrom a review standpoint: **ready to merge**.

chrisgen19:
## Fresh review pass — remaining findings

Did an independent walkthrough of the diff at `085ce91`. Earlier review items are addressed. The items below are new and actionable — ordered by impact.

---

### 1. `invalidateQueries(["recognition-cards"])` replays every loaded page (performance)

**Files:** `app/(dashboard)/dashboard/recognition/_components/recognition-form.tsx`, `app/(dashboard)/dashboard/recognition/_components/recognition-table.tsx` (not in the diff but affected by the infinite-query switch)

**Why it matters:** In TanStack Query v5, invalidating an active `useInfiniteQuery` refetches **every** loaded page in sequence, chaining `getNextPageParam` through each. After a user scrolls through 5 pages and then creates or deletes a card, the inbox page will fire 5 sequential `GET /api/recognition?cursor=...` calls. That's a real latency + Postgres cost spike on a common user action, not a rare edge case. The PR description even calls this out as an expected behavior — but it's worth a guard.

**Smallest safe fix:** In the invalidation call sites, scope to the first page:

```ts
queryClient.invalidateQueries({
  queryKey: ["recognition-cards"],
  refetchType: "active",
});
// then prune pages so the next refetch only hits page 1:
queryClient.setQueriesData<InfiniteData<FeedPage>>(
  { queryKey: ["recognition-cards"] },
  (prev) => prev ? { pages: prev.pages.slice(0, 1), pageParams: prev.pageParams.slice(0, 1) } : prev,
);
```

Or switch create/delete to an optimistic update that prepends/removes the card from page 0 directly. Either avoids N round-trips.

---

### 2. Auto-fetch cascade when the viewport exceeds the first page's rendered height

**File:** `app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx` (`RecognitionFeedInfinite` + `useIntersectionObserver({ rootMargin: "400px" })`)

**Why it matters:** The observer default `rootMargin: "400px"` means the sentinel becomes `isIntersecting: true` when it enters within 400px of the viewport. On the **dashboard widget** (smaller surface, 10-card pages), the sentinel can be in that zone **immediately on mount**, before any user scroll. Flow: mount → IO fires `isIntersecting=true` → effect fires `fetchNextPage` → page 2 loads → if cumulative height still ≤ viewport+400px, the sentinel is still intersecting → effect re-runs when `isFetchingNextPage` flips false → fires again → loop until `hasNextPage` is false. Tall cards usually mask this, but dense lists on a large monitor or the card-view `mini` mode could chain several pages on open.

**Smallest safe fix:** Track "user has scrolled at least once" or require a transition `false → true` on `isIntersecting`. Simplest:

```ts
const [hasInteracted, setHasInteracted] = useState(false);
useEffect(() => {
  const onScroll = () => setHasInteracted(true);
  window.addEventListener("scroll", onScroll, { once: true, passive: true });
  return () => window.removeEventListener("scroll", onScroll);
}, []);

useEffect(() => {
  if (hasInteracted && isIntersecting && hasNextPage && !isFetchingNextPage) {
    fetchNextPage();
  }
}, [hasInteracted, isIntersecting, hasNextPage, isFetchingNextPage, fetchNextPage]);
```

The "Load more" button still works for users who can't or don't scroll.

---

### 3. Test coverage gap: sentinel-triggered `fetchNextPage`

**Files:** `app/(dashboard)/dashboard/recognition/_components/recognition-feed.test.tsx`

**Why it matters:** The feed test mocks `useIntersectionObserver` to always return `isIntersecting: false`, so the auto-fetch-on-scroll path — the core UX of this PR — has **no test**. The existing tests exercise the "Load more" button only. If the `useEffect` gate or the observer wiring regresses, tests stay green.

**Smallest safe fix:** Add a test where the mock flips `isIntersecting: true` and assert `fetchNextPage` fires. Simplest shape:

```ts
const intersectingState = { current: false };
vi.mock("@/hooks/use-intersection-observer", () => ({
  useIntersectionObserver: () => ({ ref: () => {}, isIntersecting: intersectingState.current }),
}));

// in the test: render, await page 1, then:
act(() => { intersectingState.current = true; });
// trigger a re-render (e.g. by flipping an unrelated state)
// assert fetchMock called twice with cursor on the second call
```

---

### 4. `useIntersectionObserver`'s `enabled` option is now dead code at call sites

**Files:** `hooks/use-intersection-observer.ts`, consumer in `recognition-feed.tsx`

**Why it matters:** The PR dropped the `enabled` prop on the consumer (intentionally, to fix the reconnect race). The hook still exposes it and it's covered by a test. That's fine if you expect other call sites to use it later; if not, it's surface area that can drift out of sync with actual usage.

**Smallest safe fix:** Keep as-is (it's harmless and tested), or drop the `enabled` param from the hook signature and its test if you're confident no consumer needs it. Prefer keeping unless you're on a "minimal surface" kick.

---

### 5. Empty-state path hides `footer` even when `hasNextPage` is set (edge case)

**File:** `app/(dashboard)/dashboard/recognition/_components/recognition-feed.tsx` (`FeedLayout`)

**Why it matters:** `FeedLayout` early-returns the empty state when `cards.length === 0`, which never renders `footer`. The API's contract guarantees `nextCursor: null` when rows are empty (it derives `hasMore` from `rows.length > limit`), so this can't happen today. But the component's contract doesn't enforce that — a future caller passing zero cards + a non-null `nextCursor` would silently lose the sentinel.

**Smallest safe fix:** Either document the coupling, or render `footer` after the empty-state message. Low priority; API invariant holds.

---

### 6. `Prisma.PrismaClientValidationError` constructor drift across Prisma versions (test fragility)

**File:** `app/api/recognition/route.test.ts`

**Why it matters:** The test does `new Prisma.PrismaClientValidationError("bad cursor", { clientVersion: "x" })`. Prisma has changed the validation-error constructor signature between majors (some versions take only a string, some take an options object, some use a different discriminator). On next `bunx prisma generate` after a Prisma bump, these two tests can fail for an unrelated reason.

**Smallest safe fix:** Construct a plain object that satisfies `instanceof Prisma.PrismaClientValidationError` using `Object.setPrototypeOf`:

```ts
const makeValidationError = (msg = "bad") => {
  const e = new Error(msg);
  Object.setPrototypeOf(e, Prisma.PrismaClientValidationError.prototype);
  return e;
};
```

Or catch on `name === "PrismaClientValidationError"` in the route instead of `instanceof`. Lower coupling to Prisma internals.

---

### Non-issues I checked

- Cursor stability across same-ms `createdAt` — ✅ secondary `id desc` added.
- Reaction join scoped to trimmed page (not `+1` overfetch row) — ✅ covered in tests.
- `useInfiniteQuery` generics now `InfiniteData<FeedPage>` — ✅ clean.
- `fetchFeedPage` enforces `body.success` — ✅ clean and shared across static/infinite.
- Admin `paginated=true` / export branch gating — ✅ untouched, isolated.
- `buildFeedUrl` omits `filter=all` — matches the API default.
- `Math.floor` on clamped `limit` — ✅.
- Cursor length cap (64) — ✅ defends against accidental huge input without needing UUID-format validation.
- SSR safety in the observer hook (`typeof IntersectionObserver === "undefined"` guard) — ✅.

---

**Summary:** None are blocking for merge if infinite-scroll behaviour looks right in manual QA. **#1** (N-page replay on invalidation) and **#2** (auto-fetch cascade on short content) have the highest real-user impact. **#3** is the cheapest follow-up and closes the most important coverage gap.

chrisgen19:
## Post-fix review fixes + skeleton-shape fix applied

Commit `aa31b91`.

### Validity assessment

| # | Finding | Valid? | Applied |
|---|---|---|---|
| 1 | N-page replay on `invalidateQueries(["recognition-cards"])` | ✅ | Yes |
| 2 | Auto-fetch cascade on short content | ✅ | Yes |
| 3 | No test for sentinel-triggered `fetchNextPage` | ✅ | Yes |
| 4 | `enabled` option dead at call sites | 🟡 Minor | Skipped — hook still tests it; may be reused |
| 5 | Empty-state hides footer | 🟡 API invariant protects | Skipped |
| 6 | `Prisma.PrismaClient*Error` constructor drift | ✅ | Yes |

### What changed

**1. `lib/recognition-cache.ts`** (new) — shared `invalidateRecognitionFeed(queryClient)` that prunes each cached infinite query down to page 1 before calling `invalidateQueries`. TanStack v5's "replay all loaded pages" behaviour is now contained. Wired into:
- `recognition-form.tsx` — create / edit
- `recognition-table.tsx` — admin delete

**2. `recognition-feed.tsx`** — `RecognitionFeedInfinite` now has a `hasScrolled` gate:

```ts
useEffect(() => {
  if (hasScrolled && isIntersecting && hasNextPage && !isFetchingNextPage) {
    fetchNextPage();
  }
}, [...]);
```

First `window` scroll flips it. The "Load more" button still works unconditionally for keyboard / SR users.

**3. `recognition-feed.test.tsx`** — controllable `intersectingState` mock + new test "auto-fetches next page when sentinel intersects after a user scroll". Verifies:
- Intersecting without scroll ≠ fetch (gate works)
- Scroll event + intersecting = fetch fires with correct cursor

**4. `route.ts`** — switched from `err instanceof Prisma.PrismaClientValidationError` to `err.name === "PrismaClientValidationError"`. Same behaviour, zero dependency on Prisma's private constructor shape. Re-typed the Prisma import back to `type`-only.

### Bonus: loading-skeleton shape matches the actual card

You were right — this was a real issue, not on the reviewer's list.

The app defaults to `cardView: "physical"` which renders `RecognitionCardMini` (a large rectangular physical card with 2 columns). The loading skeleton was list-shaped (compact row with avatars + chip badges). On `/dashboard/recognition/received` and `/sent`, the skeleton-to-card swap caused a significant layout shift.

Added:
- `PhysicalCardSkeleton` — matches the mini card's 2-column layout, values panel, interaction bar footer.
- `AutoCardSkeleton` — reads `cardView` from the preferences store and renders the right variant.

Pending-state and `isFetchingNextPage` footer now both use `AutoCardSkeleton`.

### Verification

- `bun run tsc --noEmit` — clean
- `bun run biome check .` — clean
- `bun run vitest run` — **227/227** passing (14 API, 4 hook, 6 feed)

chrisgen19:
## Final pre-merge review

### CI / merge state
- **Lint + Unit tests:** ✅
- **Migrations replay:** ✅
- **E2E tests:** ✅
- **GitGuardian:** ✅
- **Vercel preview:** ✅
- **`mergeStateStatus`:** CLEAN / **`mergeable`:** MERGEABLE

### Regression check (latest commit `31db928`)

Scanned every file changed since the initial draft (11 files, +700 / -25).

| Area | Status |
|---|---|
| `GET /api/recognition` happy path (no params) | ✅ `take: 51` + slice preserves previous 50-card ceiling with `nextCursor: null` |
| Admin `?paginated=true` branch | ✅ Untouched, still returns `pagination: {page, pageSize, total, totalPages}` |
| Admin `?export=true` branch | ✅ Untouched, still 10k cap |
| Auth 401 path | ✅ Preserved |
| Reaction-summary join | ✅ Scoped to trimmed page (not `+1` overfetch row) |
| Card ordering stability | ✅ `[{createdAt: desc}, {id: desc}]` guards same-ms ties |
| Error envelope consistency | ✅ 400 for cursor-only Prisma errors, 500 for everything else |
| Existing `invalidateQueries(["recognition-cards"])` consumers | ✅ Both form and table migrated to `invalidateRecognitionFeed`; prefix still matches static + infinite |
| Non-infinite `RecognitionFeed` callers | ✅ None break — `infinite` defaults to `false`, limit defaults to 50 |
| `cardMaxWidth="max-w-3xl"` (inbox) | ✅ Now applied to skeletons too; no width snap on load |
| Skeleton shape for `cardView: "physical"` default | ✅ `AutoCardSkeleton` picks matching variant; layout shift eliminated |
| Scroll cascade on short content | ✅ `hasScrolled` gate blocks auto-fetch until first user scroll |
| "Load more" button accessibility path | ✅ Still unconditionally available |

### Remaining non-blocking observations

1. **`invalidateRecognitionFeed` updater runs against static caches too.** The `["recognition-cards"]` prefix matches the static `useQuery` too, and `setQueriesData` hands those entries to the updater. The runtime guard (`Array.isArray(prev.pages)`) correctly short-circuits, but the TypeScript generic `<InfiniteData<FeedPage>>` over-promises. A targeted unit test for this helper (two cache entries: one static, one infinite → static left untouched) would lock it in. **Not a blocker.**

2. **Scroll gate assumes `window` scroll.** Today the dashboard layout scrolls the window (confirmed in `(dashboard)/layout.tsx`). If a future layout ever wraps `main` in an overflow container, auto-fetch would silently stop triggering on inbox pages — though the "Load more" button would keep working. Worth a code comment; not a regression today.

3. **`useIntersectionObserver` `enabled` option is unused by any consumer** post-fix. The hook test covers it. Fine to leave, or drop in a small cleanup PR later.

4. **Empty-state branch hides footer.** API invariant (`nextCursor` is null whenever `data` is empty) makes this unreachable today. Same as noted in earlier rounds; still non-blocking.

### Test coverage snapshot
- API route: **14** cases (auth, limit defaults + clamping + floor, cursor forwarding + oversized cursor, error mapping 400 vs 500 with and without cursor, empty result, reaction scope).
- `useIntersectionObserver`: **4** cases (observe, callback, disabled, disconnect).
- `RecognitionFeed`: **6** cases (static, infinite first page, Load-more click, error envelope, sentinel auto-fetch with scroll gate, no-next-page).
- **Total:** 227/227 passing locally + CI.

### Verdict

**Ready to merge.** No blocking issues. CI is fully green, mergeable clean, and four review passes of hardening have been addressed. The remaining observations are minor polish items that can be cleaned up in follow-up PRs if desired.

PR #151, opened 2026-04-25
Allow OAuth-only users to set an initial password (closes #150)

### LINKED ISSUE #150: Allow OAuth-only users to set an initial password from security page
## Problem

Users who registered via Microsoft (or Google) OAuth land on `/dashboard/profile/security` and see a "Your account does not have a password login... contact an administrator" notice — they have no self-service way to add a password and unlock email/password sign-in.

The admin reset flow (`adminResetPasswordAction` in `lib/actions/user-actions.ts:288`) also bails out for these users because it only *updates* an existing `credential` account row; it cannot *create* one. So even contacting an admin doesn't solve it without code changes.

## Goal

Let an authenticated OAuth-only user set an initial password from the security page. After doing so, they can sign in with either OAuth or email + password (account linking is already enabled with `microsoft`/`google` as trusted providers in `lib/auth.ts:42`).

## Plan

1. **`lib/validations/auth.ts`** — add `setPasswordSchema` (newPassword + confirmPassword, no `currentPassword`).
2. **`lib/auth-client.ts`** — export `setPassword` from `authClient`.
3. **`app/(dashboard)/dashboard/profile/_components/set-password-form.tsx`** — new component mirroring `change-password-form.tsx`, two fields only, calls better-auth's `setPassword`.
4. **`app/(dashboard)/dashboard/profile/security/page.tsx`** — when no `credential` account exists, render `<SetPasswordForm />` (replacing the "contact administrator" notice). Keep the connected-accounts link as supporting context.
5. **`lib/auth.ts`** — add a `case "/set-password"` branch to the `after` hook to log a `PASSWORD_SET` activity. Verify `PASSWORD_SET` is part of the `ActivityAction` enum in `prisma/schema.prisma`; add it if missing (migration required).
6. **Tests** — unit test `setPasswordSchema`; sanity-check no existing tests rely on the "contact administrator" copy.
7. **Manual verification** — register via Microsoft → security page → set password → sign out → sign in via email/password. Confirm a `credential` account row is created and the regular change-password form appears on subsequent visits.

## Risks / notes

- Replaces the current "contact administrator" copy on the security page — keep the connected-accounts link visible.
- `setPassword` is session-protected by better-auth; no extra auth gate required, but keep the page's `getServerSession` redirect.
- If `PASSWORD_SET` needs to be added to `ActivityAction`, that requires a Prisma migration — flag in PR description.

chrisgen19:
## Summary

Microsoft/Google-OAuth users landing on `/dashboard/profile/security` previously saw a "contact an administrator" notice and had no self-service path to enable email/password sign-in. The admin reset action also bailed out for them (it only *updates* an existing credential row, never *creates* one). This PR closes that gap.

- Adds `SetPasswordForm` rendered on the security page when no `credential` account exists (replacing the static "contact an administrator" notice).
- Adds `setInitialPasswordAction` Server Action in `lib/actions/profile-actions.ts` — hashes via `better-auth/crypto` and inserts the credential `Account` row (`accountId === userId`, matching better-auth's internal `linkAccount` shape).
- Adds `PASSWORD_SET` to the `ActivityAction` enum and wires it into both admin activity-log views.
- Adds unit tests for the new schema and Server Action (validation, happy path, already-set, mismatch, short password).

Closes #150.

## Files

- `prisma/schema.prisma` — `PASSWORD_SET` enum value (requires `prisma migrate deploy` on prod — see Notes).
- `lib/validations/auth.ts` — `setPasswordSchema` + `SetPasswordInput` type.
- `lib/actions/profile-actions.ts` — `setInitialPasswordAction`.
- `app/(dashboard)/dashboard/profile/_components/set-password-form.tsx` — new client form.
- `app/(dashboard)/dashboard/profile/security/page.tsx` — renders `SetPasswordForm` when no credential account exists.
- `app/(dashboard)/dashboard/admin-settings/activity-logs/_components/activity-log-filters.tsx`, `…/activity-log-table.tsx` — added "Password set" label so admin views render the new action.
- `lib/validations/auth.test.ts`, `lib/actions/profile-actions.test.ts` — new tests.

## Notes / Risks

- **better-auth's `setPassword` endpoint is registered without an HTTP path** (verified in `node_modules/better-auth/dist/api/routes/update-user.mjs:188` — `createAuthEndpoint({...}, handler)` with no path string, vs. `changePassword` at line 78 which uses `"/change-password"`). Calling it from the client SDK or via `auth.api.setPassword(...)` over HTTP is not viable in this version. The Server Action mirrors `adminResetPasswordAction`'s direct-DB pattern and uses the same `hashPassword` helper better-auth uses internally, so the resulting credential row is indistinguishable from one created at signup.
- **No freshness gate on the Server Action.** better-auth's `setPassword` sits behind `sensitiveSessionMiddleware` (typically a ~5-minute fresh-session window). `setInitialPasswordAction` only requires `requireSession()`. Acceptable for an internal app, but if we want to require recent re-auth before letting an OAuth-only user create a password, we can layer that on top — say the word.
- **Prisma migration on prod.** This PR ran `prisma db push` locally (per CLAUDE.md dev workflow). When deploying to prod, ensure `prisma migrate deploy` includes the `PASSWORD_SET` enum addition, or the activity-log writes will fail at runtime.
- **No e2e test added.** The existing e2e suite doesn't cover the security page; happy to add one in a follow-up if you want it gated by CI.

## Verification (local)

- `bunx biome check` — clean (255 files).
- `bunx tsc --noEmit` — clean.
- `bun run vitest run` — 234/234 passing (28 files), including the 7 new tests.

## Test plan

- [ ] Register a fresh user via Microsoft OAuth.
- [ ] Visit `/dashboard/profile/security` → see the "Set a Password" form (not the old "contact administrator" notice).
- [ ] Submit mismatched passwords → inline error on `confirmPassword`.
- [ ] Submit a 7-char password → inline error on `newPassword`.
- [ ] Submit a valid matching pair → toast "Password set successfully", form clears, page refreshes.
- [ ] Reload `/dashboard/profile/security` → now shows the regular "Change Password" form (currentPassword required).
- [ ] Sign out, then sign in via email + password → succeeds and lands on dashboard.
- [ ] As ADMIN, visit `/dashboard/admin-settings/activity-logs` → see a `PASSWORD_SET` row labelled "Password set" with the correct actor / IP / user-agent.
- [ ] Confirm the existing `change-password` flow still works for credential-based users (regression).

## Summary by CodeRabbit

* **New Features**
  * Users without a password can set an initial password from the Security page; form enforces ≥8 chars and matching confirmation, shows success/failure feedback.

* **Bug Fixes**
  * Activity logs now display a "Password set" label for relevant entries.

* **Tests**
  * Added tests for password validation, initial-password action flows, concurrency/race conditions, and related edge cases.

* **Chores**
  * Database migrations: add PASSWORD_SET activity value and enforce unique account (user, provider) pairs.


chrisgen19:
Review findings focused on correctness/security/regressions:

## High — setting a login credential only requires any active session

- **Affected file(s):** `lib/actions/profile-actions.ts`
- **Why it matters:** `setInitialPasswordAction` creates a reusable email/password credential after only `requireSession()`. This is more sensitive than a normal profile edit: a stolen, old, or unattended OAuth session can be upgraded into a persistent password login without recent re-auth. The PR notes that Better Auth’s native set-password flow uses `sensitiveSessionMiddleware`, so this custom direct-DB path is currently weaker than the upstream security model.
- **Smallest safe fix:** Add a fresh-session/re-auth gate before hashing/creating the credential row, for example by checking a recent session timestamp if available or routing the user through a re-auth step before allowing initial password setup.

## Low — `PASSWORD_SET` logs cannot be selected in the activity-log filter

- **Affected file(s):** `app/(dashboard)/dashboard/admin-settings/activity-logs/page.tsx`
- **Why it matters:** The table/filter label maps include `PASSWORD_SET`, but the page-level `ACTIONS` whitelist still omits it. As a result, the log rows can display “Password set”, but admins cannot filter to that action, and `?action=PASSWORD_SET` is ignored.
- **Smallest safe fix:** Add `"PASSWORD_SET"` to the `ACTIONS` array used by the activity logs page.

CI is green on the latest PR run, and the PR is mergeable, but I’d treat the fresh-auth gap as a blocker because this flow creates a new login credential.

chrisgen19:
Latest review finding:

## Medium — fresh-auth gate can be bypassed if `session.session.createdAt` is invalid/missing

- **Affected file(s):** `lib/actions/profile-actions.ts`
- **Why it matters:** The stale-session check does `Date.now() - new Date(session.session.createdAt).getTime()`. If `createdAt` is malformed/undefined for any reason (adapter inconsistency, unexpected session shape, or future refactor), `getTime()` becomes `NaN`, and `sessionAgeMs > SET_PASSWORD_FRESH_AGE_MS` evaluates to `false`, which can allow password creation without a valid freshness check.
- **Smallest safe fix:** Fail closed before comparing:
  - compute the timestamp,
  - verify `Number.isFinite(createdAtMs)`,
  - return the stale-session error when invalid.

This keeps the credential-creation path secure even if session payload shape changes.

PR #182, opened 2026-07-31
feat(leaderboard): add always-visible switch to visibility settings

chrisgen19:
## Summary

Adds an **Always visible** switch to the Leaderboard Visibility panel in `/dashboard/admin-settings`. When on, the dashboard "Most Recognized" panel never locks and the monthly reveal window is bypassed. When off, the existing day-range behaviour is unchanged.

The configured reveal window is still persisted while the switch is on, so turning it back off restores the admin's schedule rather than falling back to the 1-20 defaults.

## Changes

| File | Change |
| --- | --- |
| `lib/leaderboard/visibility.ts` | `alwaysVisible` added to `LeaderboardVisibilitySettings` and `LeaderboardVisibilityState`; gate becomes `alwaysVisible \|\| inWindow`. All window dates are still computed unchanged. |
| `lib/actions/settings-actions.ts` | New `leaderboard_always_visible` app setting; read as `=== "true"`, written as a third upsert in the existing `$transaction`. |
| `app/api/recognition/stats/route.ts` | `alwaysVisible` added to the serialized `leaderboardVisibility` payload. |
| `app/(dashboard)/dashboard/_components/stats-widget.tsx` | Countdown suppressed when always-visible - there is no boundary to count toward. |
| `app/(dashboard)/dashboard/admin-settings/_components/leaderboard-visibility.tsx` | `Switch` row; both day inputs disabled and dimmed when on. Optimistic toggle with rollback on failure. |
| `app/(dashboard)/dashboard/admin-settings/page.tsx` | `initialAlwaysVisible` wired through. |

## Notes

- **No migration needed.** `AppSetting` is a key/value table; the row is created on first save. An absent row reads as `false`, so existing environments keep current behaviour.
- **Scope:** always-visible removes the lock but still shows the **previous completed month's** archived winners - not a live current-month tally.
- **Untouched:** snapshotting, the `/dashboard/leaderboard` history page, and the super-admin `previewNow` preview (a no-op while the switch is on).
- **Revert without a deploy:** `UPDATE app_settings SET value='false' WHERE key='leaderboard_always_visible';`

## Testing

- `bun run lint` - 290 files, 0 issues
- `bunx tsc --noEmit` - clean
- `bun run test` - 39 files, 403 tests passed

New coverage: 2 tests in `visibility.test.ts` (visible outside the window when on; window still honoured when off), 2 in `settings-actions.test.ts` (reads the flag; persists it alongside the days), 1 in `route.test.ts` (winners served outside the window when on).

E2E was not run - no existing spec covers this panel.

## Summary by CodeRabbit

* **New Features**
  * Added an “Always visible” option for leaderboards.
  * Leaderboards can now remain visible outside scheduled reveal windows.
  * Scheduled reveal-day settings are preserved when always-visible mode is enabled.
  * Visibility settings and status responses now include the always-visible state.
  * Countdown behavior adapts when a leaderboard is permanently visible.

* **Bug Fixes**
  * Failed visibility-toggle saves now restore the previous setting.

* **Tests**
  * Added coverage for always-visible leaderboards and settings persistence.

### DIFF
--- a/app/(dashboard)/dashboard/_components/stats-widget.tsx
+++ b/app/(dashboard)/dashboard/_components/stats-widget.tsx
@@ -43,6 +43,7 @@ interface LeaderboardVisibility {
 	revealStart: string;
 	revealEnd: string;
 	nextRevealStart: string;
+	nextMonthStart: string;
 }
 
 interface StatsData {
@@ -314,10 +315,15 @@ export function StatsWidget() {
 	const visibility = data?.data?.leaderboardVisibility ?? null;
 	// Always watch the next boundary: revealEnd if the leaderboard is currently
 	// visible (so we lock it at window-end), otherwise the next window opening.
-	// "Always visible" has no boundary at all, so there is nothing to count to.
-	const nextBoundaryIso =
-		!visibility || visibility.alwaysVisible
-			? null
+	// In always-visible mode neither is ever reached, so watch the month rollover
+	// instead — the countdown drives the refetch that advances sourceMonthKey and
+	// the archived winners on a dashboard left open past Manila midnight. Nothing
+	// is rendered from it: the countdown text only lives in LockedLeaderboard,
+	// which never mounts while visible.
+	const nextBoundaryIso = !visibility
+		? null
+		: visibility.alwaysVisible
+			? visibility.nextMonthStart
 			: visibility.visible
 				? visibility.revealEnd
 				: visibility.nextRevealStart;
--- a/app/(dashboard)/dashboard/_components/stats-widget.tsx
+++ b/app/(dashboard)/dashboard/_components/stats-widget.tsx
@@ -37,6 +37,7 @@ const STATS_STICKY_BREAKPOINT_QUERY = "(min-width: 1024px)";
 
 interface LeaderboardVisibility {
 	visible: boolean;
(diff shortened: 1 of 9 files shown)


app/(dashboard)/dashboard/_components/stats-widget.tsx
```diff
@@ -313,11 +314,13 @@ export function StatsWidget() {
...
 	// visible (so we lock it at window-end), otherwise the next window opening.
-	const nextBoundaryIso = visibility
-		? visibility.visible
-			? visibility.revealEnd
-			: visibility.nextRevealStart
-		: null;
+	// "Always visible" has no boundary at all, so there is nothing to count to.
+	const nextBoundaryIso =
+		!visibility || visibility.alwaysVisible
+			? null
```
chrisgen19: [P2] Keep a month-rollover refresh in always-visible mode

When a dashboard remains open across Asia/Manila midnight on the first day of a month, `sourceMonthKey` and the archived winners should advance, but returning `null` disables the only scheduled query invalidation. React Query's `staleTime` does not itself refetch, so the previous month's label and winners remain displayed until an unrelated invalidation, focus event, or reload; schedule a silent month-boundary refresh even though no countdown is shown.

